PR #2, opened 2021-11-24
feat(chart): initial chart files

ebaron:
Here is an initial Helm chart for Cryostat. It's mainly designed to quickly trial Cryostat in any Kubernetes cluster. As such, it's missing a lot of features of the operator, such as end-to-end encryption, authentication, persistent storage. Some of these should be not too difficult to add later.

It should work with a variety of network setups:
- ClusterIP services, and `kubectl port-forward`
- NodePort services
- LoadBalancer services
- Ingress
- Routes (on OpenShift)

I've tested it out manually using Minikube and CodeReady Containers. If new to Helm, this is a good resource to learn the basics: https://helm.sh/docs/intro/.

To install:
`helm install <name> ./cryostat/`

To uninstall:
`helm uninstall <name>`

To render the YAML files that would be created during installation:
`helm template <name> ./cryostat/`

The `install` and `template` commands allow you to override values to test different scenarios without modifiying the values.yaml directly, e.g. `--set core.service.type=NodePort,grafana.service.type=NodePort` or `--set core.route.enabled=true,grafana.route.enabled=true`.


cryostat/values.yaml
```diff
@@ -0,0 +1,152 @@
...
+  #   memory: 128Mi
+  securityContext: {}
+  # capabilities:
+  #   drop:
+  #   - ALL
+  # readOnlyRootFilesystem: true
+  # runAsNonRoot: true
+  # runAsUser: 1000
+
+grafana:
```
heubeck, line 51: shall it be possible, to use an existing grafana instance as well?

cryostat/Chart.yaml
```diff
@@ -0,0 +1,51 @@
+apiVersion: v2
+name: cryostat
+description: Securely manage JFR recordings for your containerized Java workloads
+
+# A chart can be either an 'application' or a 'library' chart.
```
heubeck, line 5: guess you'd like to remove the default comments.

cryostat/values.yaml
ebaron: Not at first, but I could see that being a useful feature. I've filed #3 to make a note of this (linking to our operator since this would also be useful there).

cryostat/Chart.yaml
ebaron: Definitely, thanks for the reminder!

cryostat/templates/NOTES.txt
```diff
@@ -0,0 +1,84 @@
...
+{{- $portForwards = append $portForwards "8081:$GRAFANA_CONTAINER_PORT" }}
+{{- end }}
+
+{{- if not (empty $envVars) }}
+  kubectl -n {{ .Release.Namespace }} set env deploy --containers={{ .Chart.Name }} {{ include "cryostat.fullname" . }} {{ join " " $envVars }}
+{{- end }}
+
+{{- if not (empty $portForwards) }}
+
+{{ $listNum }}. Forward local ports to the application's pod:
```
andrewazores, line 63: I tried running these steps from the `NOTES` that were printed and ran into an error on this `kubectl port-forward` step because I had done it too quickly and my Pod wasn't running yet. Would it make sense to add a step here with an ex. `kubectl wait --for=condition=available --timeout=60s` to wait for the deployment to be stable/ready before moving on?

heubeck (approved):
Thank you.
Will you serve this chart from a http repo, e.g. by using GH pages?

cryostat/templates/NOTES.txt
ebaron: Good suggestion! I've done this now

ebaron:
> Thank you. Will you serve this chart from a http repo, e.g. by using GH pages?

We still have to explore release options for this chart, but serving it on our GH page could be a good option. This PR references images from our latest development branch. Once we release our 2.1 version of Cryostat, we would like to release a Helm chart to go with it. It would also be nice if we could release a separate chart version pointing the latest development branch as well. I've created #8 to explore this further.

heubeck:
> > Thank you. Will you serve this chart from a http repo, e.g. by using GH pages?
> 
> We still have to explore release options for this chart, but serving it on our GH page could be a good option. This PR references images from our latest development branch. Once we release our 2.1 version of Cryostat, we would like to release a Helm chart to go with it. It would also be nice if we could release a separate chart version pointing the latest development branch as well. I've created #8 to explore this further.

Sounds great.
A nice setup is e.g. here: https://github.com/bitnami-labs/sealed-secrets
They are creating GH releases with the packaged chart using GH actions, and host the helm repo index with GH actions.

andrewazores (approved):

ebaron:
> Sounds great. A nice setup is e.g. here: https://github.com/bitnami-labs/sealed-secrets They are creating GH releases with the packaged chart using GH actions, and host the helm repo index with GH actions.

Thanks for the tip, I'll take a look!

Issue #22, opened 2022-05-10
Use Markdown syntax for code in NOTES.txt


ebaron:
For example, in the OpenShift Console, NOTES.txt is interpreted as Markdown. This can cause some variable names to be incorrect.

PR #23, opened 2022-05-10
fix(notes): use code blocks for shell commands

ebaron:
This also fixes a typo in one of the branch conditions.

Fixes: #22

### DIFF
--- a/cryostat/Chart.yaml
+++ b/cryostat/Chart.yaml
@@ -4,7 +4,7 @@ description: Securely manage JFR recordings for your containerized Java workload
 
 type: application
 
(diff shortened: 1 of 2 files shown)


ebaron:
@andrewazores You can install the chart from this PR with `helm install cryostat ./cryostat`. Then in the OpenShift Console, open the Developer perspective and Topology view. You should see the Helm Chart there, click on the "HR" component and it should open the "Release notes" tab like this:

Screenshot 2022-05-10 at 14-33-01 Topology · Red Hat OpenShift Container Platform

andrewazores (approved):

PR #36, opened 2022-09-30
Allow SSL Proxied EnvVars to be enabled outside of Ingress Configuration

hgranillo:
Hi cryostat helm chart maintainers!

This PR adds the ability to set up a parameter `core.sslProxied` in order to enable all the "SSL Proxied" environment variables outside of ingress/route tls configuration.

The main reason behind this is that when you are offloading at the loadbalancer and not at the ingress you have no way currently to enable this settings in the current version of the chart.

I've tested this in my current deployment of cryostat and seems to work fine, I stopped receiving errors related to connection errors.

### DIFF
--- a/cryostat/templates/deployment.yaml
+++ b/cryostat/templates/deployment.yaml
@@ -44,7 +44,7 @@ spec:
           - name: CRYOSTAT_PROBE_TEMPLATE_PATH
             value: /opt/cryostat.d/probes.d
           - name: CRYOSTAT_EXT_WEB_PORT
-            value: "{{ if (or (and .Values.core.route.enabled .Values.core.route.tls.enabled) (and .Values.core.ingress.enabled .Values.core.ingress.tls)) }}443{{ else }}80{{ end }}"
+            value: "{{ if (or (and .Values.core.route.enabled .Values.core.route.tls.enabled) (and .Values.core.ingress.enabled .Values.core.ingress.tls) (.Values.core.sslProxied)) }}443{{ else }}80{{ end }}"
           - name: CRYOSTAT_WEB_HOST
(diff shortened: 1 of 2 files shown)


cryostat/templates/deployment.yaml
```diff
@@ -54,7 +54,7 @@ spec:
           - name: GRAFANA_DATASOURCE_URL
             value: http://127.0.0.1:8080
           - name: GRAFANA_DASHBOARD_URL
-            value: "{{ if .Values.grafana.ingress.enabled }}http{{ if .Values.grafana.ingress.tls }}s{{ end }}://{{ with index .Values.grafana.ingress.hosts 0 }}{{ .host }}{{ end }}{{ end }}"
+            value: "{{ if .Values.grafana.ingress.enabled }}http{{ if (or (.Values.grafana.ingress.tls) (.Values.core.sslProxied)) }}s{{ end }}://{{ with index .Values.grafana.ingress.hosts 0 }}{{ .host }}{{ end }}{{ end }}"
```
ebaron, line 57: I'm not sure why someone would do this, but it's possible that the Grafana service could be exposed in a different way than the core service (e.g. an Ingress with no TLS). To cover this case, how about also adding a `grafana.sslProxied` as well? This way the core and grafana configurations can be kept independent.

ebaron (commented):
Hi @hgranillo, thank you so much for the contribution! This is something I hadn't considered and is an example of why it's great to hear from users on how they're deploying Cryostat.

I think the approach you've taken makes sense, I just have one suggestion below to account for an unusual use case.

cryostat/templates/deployment.yaml
hgranillo: Hi @ebaron, I agree with comment and I've updated my PR accordingly.

Thank you for your comments!

ebaron (approved):
Looks good! Thanks again for contributing this fix!

PR #42, opened 2022-10-25
chore(versions): set version tags for 2.2.0 release point

andrewazores:
Related to #1118

### DIFF
--- a/cryostat/values.yaml
+++ b/cryostat/values.yaml
@@ -7,7 +7,7 @@ core:
     ## @param core.image.pullPolicy Image pull policy for the main Cryostat container image
     pullPolicy: Always
     ## @param core.image.tag Tag for the main Cryostat container image
-    tag: "2.2"
+    tag: "2.2.0-beta1"
   service:
     ## @param core.service.type Type of Service to create for the Cryostat application
     type: ClusterIP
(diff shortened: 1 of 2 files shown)


cryostat/values.yaml
```diff
@@ -66,7 +66,7 @@ grafana:
     ## @param grafana.image.pullPolicy Image pull policy for the Grafana container image
     pullPolicy: Always
     ## @param grafana.image.tag Tag for the Grafana container image
-    tag: "latest"
+    tag: "v2.2"
```
ebaron, line 69: I think this one should be `cryostat-v2.2` for the builds from Quay.

cryostat/values.yaml
```diff
@@ -121,7 +121,7 @@ datasource:
     ## @param datasource.image.pullPolicy Image pull policy for the JFR Data Source container image
     pullPolicy: Always
     ## @param datasource.image.tag Tag for the JFR Data Source container image
-    tag: "latest"
+    tag: "2.2"
```
ebaron, line 124: Same question as above with Cryostat.

andrewazores: I explained some context for these choices here:

https://github.com/cryostatio/cryostat-operator/pull/480

whichever way we decide, I will update the relevant PR(s).

ebaron (approved):

Issue #43, opened 2022-11-03
Set SecurityContext defaults for restricted PSA


ebaron:
It would be useful to set this by default to not require the user to specify it themselves.

PR #44, opened 2022-11-03
fix(security): default to restricted pod security

ebaron:
Fixes: #43


tthvo (approved):
Looks great to me! Worked as expected. Though, we should add a newline in `values.schema.json`.

Follow-up: Since we set `seccomp` here, it is not compatible with OCP < 4.11, so upstream only works with OCP >= 4.11? Should we start adding `kubeVersion` to `Charts.yaml` https://helm.sh/docs/topics/charts/#the-chartyaml-file?

ebaron:
> Looks great to me! Worked as expected. Though, we should add a newline in `values.schema.json`.
> 
> Follow-up: Since we set `seccomp` here, it is not compatible with OCP < 4.11, so upstream only works with OCP >= 4.11? Should we start adding `kubeVersion` to `Charts.yaml` https://helm.sh/docs/topics/charts/#the-chartyaml-file?

Right, in the downstream version at https://github.com/openshift-helm-charts/charts, I plan to patch out `seccompProfile`. We should add `kubeVersion`, I filed an issue for it here: https://github.com/cryostatio/cryostat-helm/issues/45

Issue #57, opened 2023-04-19
[Bug] Cryostat can't discover its own namespace


ebaron:
### Current Behavior

Cryostat installed with the Helm Chart no longer sees any applications due to not setting `CRYOSTAT_K8S_NAMESPACES`.

### Expected Behavior

Applications in Cryostat's namespace should be discovered.

### Steps To Reproduce

1. Install Cryostat latest/2.3.0-SNAPSHOT using the Helm Chart
2. Check list of targets

### Anything else?

![Screenshot 2023-04-19 at 18-23-35 Dashboard](https://user-images.githubusercontent.com/4326090/233214023-3bca93f7-db75-48f7-8534-1bfa1a42a119.png)

PR #58, opened 2023-04-19
fix(deploy): set CRYOSTAT_K8S_NAMESPACES

ebaron:
Fixes: #57


tthvo (approved):
Looks good! Seeing discovered targets now :))

PR #71, opened 2023-06-01
feat(ci): chart release workflow

### LINKED ISSUE #8: [Story] Explore release options
Look into way we can publish releases to coincide with Cryostat project releases, and also snapshot releases of the latest development branch.

Relevant GH action:
https://helm.sh/docs/howto/chart_releaser_action/

### LINKED ISSUE #76: [Task] Switch to GitHub's generated release notes
The Release Drafter action creates draft releases, which will probably conflict with the releases that the Chart Releaser Action tries to make. I think it should actually be possible to substitute GitHub's built-in release notes generator for this action. We add a file to customize the format of the release notes, using labels as we do currently:
https://docs.github.com/en/repositories/releasing-projects-on-github/automatically-generated-release-notes#configuring-automatically-generated-release-notes

The Chart Releaser Action uses the GitHub Release API to generate release notes if they're not provided. Unfortunately this API doesn't provide a way to specify the previous tag to compare against. It does have a separate API to do this and return the release notes as JSON: https://docs.github.com/en/rest/releases/releases?apiVersion=2022-11-28#generate-release-notes-content-for-a-release

We can call this API inside the workflow which should already have the GitHub CLI installed:
```
# GitHub CLI api
# https://cli.github.com/manual/gh_api

gh api \
  --method POST \
  -H "Accept: application/vnd.github+json" \
  -H "X-GitHub-Api-Version: 2022-11-28" \
  /repos/OWNER/REPO/releases/generate-notes \
  -f tag_name='v1.0.0' \
 -f target_commitish='main' \
 -f previous_tag_name='v0.9.2' \
 -f configuration_file_path='.github/custom_release_config.yml' 
```

We can use jq to grab the body from the response and output it to a file. Then create a config.yaml with `release-notes-file: /path/to/release-notes`[1]. Then pass the config file to the Chart Releaser Action like this: https://github.com/helm/chart-releaser-action#example-using-custom-config

We might be able to use [this action](https://github.com/PaulHatch/semantic-version) to quickly compute the previous semver tag. If not it should be easy to do ourselves.

[1] https://github.com/helm/chart-releaser/blob/fce24e7688b7dd9b6091ff2e84d1eaabce8de1ee/pkg/config/config.go#LL61C45-L61C63

_Originally posted by @ebaron in https://github.com/cryostatio/cryostat-helm/issues/71#issuecomment-1585073797_

tthvo:
Fixes #8 
Fixes #76 
Depends on #73 (i.e. restructure chart directory)
Depends on https://github.com/cryostatio/cryostatio.github.io/pull/132

Release CI for helm charts. Release drafter is replaced with Github automated release note generations to avoid conflict with helm-releaser-actions. See #76.


ebaron (commented):
@tthvo thanks for taking this on! I've been trying to think of ways we can make this work with the Chart Releaser Action. What about making the workflow manual only?
https://docs.github.com/en/actions/managing-workflow-runs/manually-running-a-workflow

It should allow us to run the workflow on a branch (when we're ready to release), and then leave the rest up to the Chart Releaser Action?

tthvo:
> What about making the workflow manual only?
> https://docs.github.com/en/actions/managing-workflow-runs/manually-running-a-workflow

> It should allow us to run the workflow on a branch (when we're ready to release), and then leave the rest up to the Chart Releaser Action?

Oh yes manual trigger sounds good! I am just concerned about these requirements: https://github.com/helm/chart-releaser-action#pre-requisites

It requires `gh-pages` branch to work while upstream doc build is done on `main`...

ebaron:
> It requires `gh-pages` branch to work while upstream doc build is done on `main`...

Do you mean the release notes drafter?

tthvo:
> Do you mean the release notes drafter?

Oh no i mean our upstream website here: https://github.com/cryostatio/cryostatio.github.io. I remember we want to host the chart repository under `cryostat.io/helm-charts` right?

ebaron:
> > Do you mean the release notes drafter?
> 
> Oh no i mean our upstream website here: https://github.com/cryostatio/cryostatio.github.io. I remember we want to host the chart repository under `cryostat.io/helm-charts` right?

Oh right. Are we able to add the `gh-pages` branch of this repo as a submodule to the website's main repo?

tthvo:
> Are we able to add the gh-pages branch of this repo as a submodule to the website's main repo?

Oh wait now that it makes sense about the submodule. I think its really possible. 

In this PR, I am doing updates manually. Do u think its easier this way or submodule is better?

ebaron:
> In this PR, I am doing updates manually. Do u think its easier this way or submodule is better?

Sorry, what updates are you referring to?

Also, I thought of another problem. The Release Drafter action creates draft releases, which will probably conflict with the releases that the Chart Releaser Action tries to make. I think it should actually be possible to substitute GitHub's built-in release notes generator for this action. We add a file to customize the format of the release notes, using labels as we do currently:
https://docs.github.com/en/repositories/releasing-projects-on-github/automatically-generated-release-notes#configuring-automatically-generated-release-notes

The Chart Releaser Action uses the GitHub Release API to generate release notes if they're not provided. Unfortunately this API doesn't provide a way to specify the previous tag to compare against. It does have a separate API to do this and return the release notes as JSON: https://docs.github.com/en/rest/releases/releases?apiVersion=2022-11-28#generate-release-notes-content-for-a-release

We can call this API inside the workflow which should already have the GitHub CLI installed:
```
# GitHub CLI api
# https://cli.github.com/manual/gh_api

gh api \
  --method POST \
  -H "Accept: application/vnd.github+json" \
  -H "X-GitHub-Api-Version: 2022-11-28" \
  /repos/OWNER/REPO/releases/generate-notes \
  -f tag_name='v1.0.0' \
 -f target_commitish='main' \
 -f previous_tag_name='v0.9.2' \
 -f configuration_file_path='.github/custom_release_config.yml' 
```

We can use jq to grab the body from the response and output it to a file. Then create a config.yaml with `release-notes-file: /path/to/release-notes`[1]. Then pass the config file to the Chart Releaser Action like this: https://github.com/helm/chart-releaser-action#example-using-custom-config

We might be able to use [this action](https://github.com/PaulHatch/semantic-version) to quickly compute the previous semver tag. If not it should be easy to do ourselves.

[1] https://github.com/helm/chart-releaser/blob/fce24e7688b7dd9b6091ff2e84d1eaabce8de1ee/pkg/config/config.go#LL61C45-L61C63

tthvo:
> Sorry, what updates are you referring to?

Sorry I am referring to updates to this `index.yaml` file here: https://github.com/cryostatio/cryostatio.github.io/pull/132. Currently, this PR will listen for a new release and patch that file with new chart version.

> Also, I thought of another problem....

Ah yes! I will look into the direction then if we go with submodule + Release Action then.

ebaron:
> > Sorry, what updates are you referring to?
> 
> Sorry I am referring to updates to this `index.yaml` file here: [cryostatio/cryostatio.github.io#132](https://github.com/cryostatio/cryostatio.github.io/pull/132). Currently, this PR will listen for a new release and patch that file with new chart version.

The submodule seems like it would avoid you having to do this. I would concentrate first on getting the workflow working in this repository. Once that's done we can see how to integrate it into the website. Hopefully it's as easy as it sounds.

> 
> > Also, I thought of another problem....
> 
> Ah yes! I will look into the direction then if we go with submodule + Release Action then.

I filed #76 since it's tangential, but releated work.

tthvo:
Thanks I will get back to this after PTO then. I think it should be easy enough to have `index.yaml` hosted on `cryostat.io/helm-charts` :))

tthvo:
Release: https://github.com/tthvo/cryostat-helm/releases

Gh-pages update: https://github.com/tthvo/cryostat-helm/tree/gh-pages (see `index.yaml`).

Action Run: https://github.com/tthvo/cryostat-helm/actions/runs/5350680838/jobs/9703606956

tthvo:
I got some issue when testing the release note on my fork as nothing was reported. However, when testing on upstream, the release note is generated correctly (i.e. using my upstream branch as main does not have config file):

```bash
$ gh api --method POST -H "Accept: application/vnd.github+json" /repos/cryostatio/cryostat-helm/releases/generate-notes -f tag_name='v0.4.0' -f target_commitish='helm-release' -f previous_tag_name='v0.3.0' | jq -r .body
<!-- Release notes generated using configuration in .github/release.yml at helm-release -->

## What's Changed
### 🚀 Features
* feat(deploy): minimal parameter for deploying Cryostat without grafana and datasource by @tthvo in https://github.com/cryostatio/cryostat-helm/pull/66
* feat(ci): workflow to lint and test charts  by @tthvo in https://github.com/cryostatio/cryostat-helm/pull/73
### 🔧 Fixes
* fix(chart): consider pre-release versioning by @ebaron in https://github.com/cryostatio/cryostat-helm/pull/59
* fix(deploy): set CRYOSTAT_K8S_NAMESPACES by @ebaron in https://github.com/cryostatio/cryostat-helm/pull/58
* fix(secret): use release name in database secret by @tthvo in https://github.com/cryostatio/cryostat-helm/pull/65
* fix(note): fix incorrect service field in NOTES.txt by @tthvo in https://github.com/cryostatio/cryostat-helm/pull/68
* fix(ci): increase timeout for chart testing process by @tthvo in https://github.com/cryostatio/cryostat-helm/pull/74
### ⚙️ Maintenance
* chore(github): add pr + issue templates + add release drafter by @maxcao13 in https://github.com/cryostatio/cryostat-helm/pull/56
* ci(mergify): update config for latest release branch by @tthvo in https://github.com/cryostatio/cryostat-helm/pull/55

## New Contributors
* @maxcao13 made their first contribution in https://github.com/cryostatio/cryostat-helm/pull/56

**Full Changelog**: https://github.com/cryostatio/cryostat-helm/compare/v0.3.0...v0.4.0

```

tthvo:
Also, i wonder if we could initialize the gh-pages branch with an intial `index.yaml` that contains previous releases? New chart release will just patch that file with new release information.

I already generated one here: https://github.com/cryostatio/cryostatio.github.io/pull/132/files#diff-09d0d2fba18742bdcc0d3f6a345da7e93c6522d1fdf5bcb02a9ed5d47b05219c

ebaron:
> Also, i wonder if we could initialize the gh-pages branch with an intial `index.yaml` that contains previous releases? New chart release will just patch that file with new release information.
> 
> I already generated one here: [cryostatio/cryostatio.github.io@95076a5#diff-09d0d2fba18742bdcc0d3f6a345da7e93c6522d1fdf5bcb02a9ed5d47b05219c](https://github.com/cryostatio/cryostatio.github.io/commit/95076a58157df20f5cd4390da4db654024dc6d32#diff-09d0d2fba18742bdcc0d3f6a345da7e93c6522d1fdf5bcb02a9ed5d47b05219c)

That should be fine.

tthvo:
Nice thanks! Created and cleaned up for `gh-pages` branch here: https://github.com/cryostatio/cryostat-helm/tree/gh-pages

I will rebase it after this PR just to mark where it all started.

.github/workflows/helm-release.yml
```diff
@@ -0,0 +1,47 @@
...
+          git config user.email "$GITHUB_ACTOR@users.noreply.github.com"
+      - name: Get release version
+        id: release-version
+        run: |
+          echo "tag=v$(yq .version charts/cryostat/Chart.yaml)" >> $GITHUB_OUTPUT
+      - name: Get previous version
+        id: previous-version
+        uses: WyriHaximus/github-action-get-previous-tag@v1
+        with:
+          fallback: 0.1.0
```
ebaron, line 28: This action doesn't seem to consider the current branch/tag, so if for some reason we needed to release a 0.2.x after 0.3.0 has been released, it will return 0.3.0 as the tag to compare against. We don't plan on doing this now, but it might cause problems in the future.

Another problem is that while it accepts a pattern to match tags with, it uses it as a prefix only. So this could match things like `0.4.0-BETA` for example, and there's no way to avoid it.

We could avoid the second problem by calling `git for-each-ref` directly:
`git for-each-ref --sort=-creatordate --count 1 --format="%(refname:short)" "refs/tags/v[0-9]*.[0-9]*.[0-9]*"`.

Or we could handle both issues with something like this:
```bash
# Filter tags by regex, combine with the Chart version, sort by version number, and output the preceeding version.
version="$(yq .version charts/cryostat/Chart.yaml)"; \
{ echo v${version}; git tag -l | grep -E '^v[0-9]\.+[0-9]+\.[0-9]+$'; } | \
sort -V | grep -B1 "${version}" | head -n1
```
What do you think?

tthvo: Ahh understood! Sounds good! Updated just now thanks!

tthvo: Also, regex should instead be `^v[0-9]+\.[0-9]+\.[0-9]+$`?

ebaron (approved):
Looks good! Thanks a lot @tthvo, excellent work as always!

.github/workflows/helm-release.yml
ebaron: Ah sorry, typo.

PR #74, opened 2023-06-05
fix(ci): increase timeout for chart testing process

tthvo:
Related to #73 

This increases the timeout for the testing job.

### DIFF
--- a/ct.yaml
+++ b/ct.yaml
@@ -3,3 +3,4 @@ chart-dirs:
   - charts
 check-version-increment: false
 validate-maintainers: false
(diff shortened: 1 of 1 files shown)


tthvo:
@ebaron  Hey, I think it might just be the timeout issue like you said. 

Same thing happened here: https://github.com/cryostatio/cryostat-helm/actions/runs/5182531526/jobs/9339419644 (run for 8ms then failed).

ebaron (approved):
Thanks for investigating this!

PR #105, opened 2023-12-08
ci(release): call submodule update from release workflow

ebaron:
Won't fix the problem for this release unfortunately.

Fixes: #104


andrewazores:
> Won't fix the problem for this release unfortunately.

We could delete the release and re-run the action, I think

ebaron:
@Mergifyio backport gh-pages

andrewazores:
Ah, right. I think just reverting that commit so the workflow action can redo it makes sense.

PR #109, opened 2023-12-08
chore(version): prepare for 2.4.1 development

andrewazores:
Related to cryostatio/cryostat#1695


andrewazores:
Done: https://github.com/cryostatio/cryostat/releases/tag/v2.4.1-dev

andrewazores:
I guess there needs to be a new snapshot build of the Cryostat container so that it picks that tag up and uses that in the health response.

andrewazores:
Maybe we want to apply that pretty generally to most/all of our CI actions - if it is triggered by a push to some branch(es), it should also be triggered by tags being pushed?

ebaron:
> Maybe we want to apply that pretty generally to most/all of our CI actions - if it is triggered by a push to some branch(es), it should also be triggered by tags being pushed?

I did something like this, but my regex was too restrictive:
https://github.com/cryostatio/cryostat/blob/af1d386b54380d57744af4f4e270432ef78e33fc/.github/workflows/push-ci.yml#L15

Perhaps it should just be:
```yaml
on:
  push:
    tags:
```
if that's valid syntax.

andrewazores:
I don't think it's totally clear here: https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#onpushbranchestagsbranches-ignoretags-ignore

It says that this "accepts" glob patterns, so they might be optional, but it doesn't explicitly say and there are no examples without provided patterns.

ebaron:
The regex is optional according to this: https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#patterns-to-match-branches-and-tags

andrewazores:
Cryostat 2.4.1-snapshot upstream image is finally up on quay.

```bash
$ https :8181/health
HTTP/1.1 200 OK
content-encoding: gzip
content-length: 139
content-type: application/json

{
    "cryostatVersion": "v2.4.1-dev-5-gd5f181d0",
    "dashboardAvailable": false,
    "dashboardConfigured": false,
    "datasourceAvailable": false,
    "datasourceConfigured": false,
    "reportsAvailable": true,
    "reportsConfigured": false
}
```

ebaron (approved):

Issue #110, opened 2023-12-18
[Story] Deploy Cryostat 3.0


andrewazores:
### Describe the feature

The chart should be updated to deploy the development builds of Cryostat 3.0.

- [x] #113
- [x] #114
- [ ] #115
- [x] #116
- [x] #117

andrewazores:
Done other than #115, which will be moved to the next release as a potential item to pick up.

PR #111, opened 2023-12-18
feat(cryostat): deploy cryostat 3.0

### LINKED ISSUE #113: [Task] Deploy Cryostat 3.0 with cryostat-storage, cryostat-db

andrewazores:
Related to #110
Fixes #113
Depends on https://github.com/cryostatio/cryostat3/pull/209
Depends on https://github.com/cryostatio/cryostat-storage/pull/1

This is a very rough first pass at getting Cryostat 3.0.0-snapshot deployed. A cryostat-db instance and cryostat-storage instance are deployed alongside the Cryostat container as part of the same Pod.

These things are still TODO and I intend to tackle them as follow-up PRs either into this same feature branch, or into the upstream base branch I'm targeting:
- [x] ~~figure out minio root user config issue. Works on OpenShift (`crc`), complains on `kind`: `Error: container has runAsNonRoot and image will run as root (pod: "cryostat-74cd7858bc-sgj5d_default(7e38eae9-9189-441c-816c-4bd4db57f219)", container: cryostat-s3)`~~ partially solved in latest changes and with https://github.com/cryostatio/cryostat-storage . Probably needs more robust checks and will affect permissions for mounted volumes for the next point down:
- [x] add persistent volumes for db and storage
- [x] configure credentials and Secret for the db and s3 connections
- [ ] deploy an oauth2-proxy and populate it with an `htpasswd` providing Basic auth using the existing configuration options #114 #116
- [ ] implement an option for OpenShift auth that replaces the oauth2-proxy with an openshift-oauth-proxy: requires https://github.com/andrewazores/oauth-proxy/tree/go-1.22 #117
- [ ] grafana deployment is currently broken, but this is going to change anyway with the auth proxy setups so that the auth proxy is the only directly exposed component. In the case where no auth is configured, the oauth2 proxy can be deployed with the "SKIP_ROUTES" env var set to a wildcard, so that the overall deployment still looks the same but the proxy just passes all requests. #114
- [ ] move cryostat-storage into its own Deployment with its own auth proxy #115


charts/cryostat/README.md
```diff
@@ -32,6 +32,19 @@ A Helm chart for deploying [Cryostat](https://cryostat.io/) on Kubernetes and Op
...
+| `storage.image.repository` | Repository for the storage container image                                                                                                                                                                                                                                                                    | `quay.io/cryostat/cryostat-storage` |
+| `storage.image.pullPolicy` | Image pull policy for the storage container image                                                                                                                                                                                                                                                             | `Always`                            |
+| `storage.image.tag`        | Tag for the storage container image                                                                                                                                                                                                                                                                           | `2023-12-19`                        |
```
ebaron, line 43: I think the README/schema need to be regenerated after changing the default to `latest`.

charts/cryostat/templates/deployment.yaml
```diff
@@ -34,69 +36,61 @@ spec:
...
-          - name: CRYOSTAT_HIBERNATE_DIALECT
-            value: org.hibernate.dialect.H2Dialect
-          - name: CRYOSTAT_JDBC_USERNAME
-            value: {{ include "cryostat.fullname" . }}
-          - name: CRYOSTAT_JDBC_PASSWORD
-            value: {{ include "cryostat.fullname" . }}
-          {{- end }}
-          - name: CRYOSTAT_JMX_CREDENTIALS_DB_PASSWORD
+          - name: QUARKUS_S3_AWS_REGION
+            value: us-east-1
```
ebaron, line 70: Is this just a dummy value for now? Does it do something with SeaweedFS?

charts/cryostat/templates/deployment.yaml
```diff
@@ -34,69 +36,61 @@ spec:
...
-          {{- if (.Values.authentication).basicAuth.enabled }}
-            value: io.cryostat.net.BasicAuthManager
-          {{- else }}
-            value: io.cryostat.net.NoopAuthManager
+          - name: AWS_SECRET_ACCESS_KEY
+            valueFrom:
+              secretKeyRef:
+                name: {{ printf "%s-storage-secret-key" .Release.Name }}
+                key: SECRET_KEY
+                optional: false
```
ebaron, line 88: For my own info, why are there two pairs of these environment variables? (e.g. `QUARKUS_S3_AWS_CREDENTIALS_STATIC_PROVIDER_SECRET_ACCESS_KEY` and `AWS_SECRET_ACCESS_KEY`).

ebaron (commented):
Just a few comments so far, mainly for my own knowledge.

charts/cryostat/templates/deployment.yaml
andrewazores: It doesn't actually do anything when using SeaweedFS, or probably when using any other "self-hosted" kind of S3 provider, but the SDK client wants this variable to be set to something - presumably because the actual AWS S3 requires it.

charts/cryostat/templates/deployment.yaml
andrewazores: I need to double-check that myself, actually. I think the `QUARKUS_` one is what is actually needed on the Cryostat side, and the `AWS_` one was needed for Minio previously? Or perhaps the SDK client does actually need both, for some reason.

charts/cryostat/templates/deployment.yaml
andrewazores: https://docs.quarkiverse.io/quarkus-amazon-services/dev/amazon-s3.html#_configuring_s3_clients

> quarkus.s3.aws.region - It’s required by the client, but since you’re using a local S3 instance you can pick any valid AWS region.

charts/cryostat/templates/deployment.yaml
andrewazores: I think this was probably just a configuration hold-over:

https://docs.quarkiverse.io/quarkus-amazon-services/dev/amazon-s3.html#quarkus-amazon-s3_quarkus.s3.aws.credentials.type

If using the `env-variable` credentials type then the `AWS_SECRET_ACCESS_KEY` would be checked. Since in any case this configuration is going to expose the credentials to the container via environment variables, just using `static` with the `QUARKUS_`-prefixed pair is simpler.

charts/cryostat/templates/deployment.yaml
andrewazores: It would also be possible to hardcode a default value for this property in https://github.com/cryostatio/cryostat3/blob/main/src/main/resources/application.properties , so that end users can still override it if necessary but have a fallback if they don't, but I'm not sure if that's worthwhile. In the end if the user is going to choose some other S3 provider than the one we ship, they'll have to change all these related configuration knobs anyway.

ebaron (approved):
Looks good! Thanks for doing this

charts/cryostat/templates/deployment.yaml
andrewazores: Looks like this was actually used for something after all: https://github.com/cryostatio/cryostat-helm/issues/120

Issue #116, opened 2024-01-18
[Task] Reimplement Basic auth via oauth2-proxy htpasswd config


andrewazores:
References:
- https://github.com/cryostatio/cryostat3/blob/997f1419e77a5776b7e92368069ca007b86cb287/smoketest/compose/auth_proxy.yml#L41
- https://github.com/cryostatio/cryostat3/blob/997f1419e77a5776b7e92368069ca007b86cb287/smoketest/compose/auth_proxy_htpasswd#L1
- https://oauth2-proxy.github.io/oauth2-proxy/docs/configuration/overview (search for `htpasswd`)

PR #118, opened 2024-01-18
feat(cryostat): re-implement  basic auth via oauth2-proxy htpasswd config

aali309:
fixes: #116


charts/cryostat/templates/auth_proxy.yaml
```diff
@@ -0,0 +1,19 @@
+version: "3"
+services:
```
andrewazores, line 2: I don't think this will work... ? This is a docker-compose definition, but what we need in the helm chart are, well, helm resources, which are templated versions of Kubernetes resources.

Take a look at this file in my PR for example: https://github.com/cryostatio/cryostat-helm/pull/111/files#diff-45046576393413b4292b320376e53136d68615d83d721b3fde4970e5985098a5

andrewazores: I think you should also rebase on top of #111 , otherwise you'll be trying to put an auth proxy in front of Cryostat 2.4, which is quite a different setup (no database, no storage container, etc.)

andrewazores:
Not sure why it's showing so many conflicts - just try a `git fetch upstream && git merge rebase -i upstream/main` and drop any commits from my PR that git doesn't realize are already incorporated in the base branch.

.github/workflows/helm-release.yml
```diff
@@ -39,14 +39,13 @@ jobs:
             -f target_commitish="${GITHUB_REF_NAME}" \
             -f previous_tag_name="${{ steps.previous-version.outputs.tag }}" | jq -r .body > charts/cryostat/release-notes.md
       - name: Run chart-releaser
-        uses: helm/chart-releaser-action@v1.5.0
+        uses: helm/chart-releaser-action@v1.6.0
```
andrewazores, line 42: I think this change should be undone and split out into a separate PR (against `main`)

charts/cryostat/templates/deployment.yaml
```diff
@@ -190,6 +190,14 @@ spec:
...
             {{- toYaml .Values.storage.securityContext | nindent 12 }}
         {{- if not .Values.minimal }}
+        - name: {{ printf "%s-%s" .Chart.Name "auth-proxy" }}
+          securityContext:
+            {{- toYaml .Values.authProxy.securityContext | nindent 12 }}
+          image: "{{ .Values.authProxy.image.repository }}:{{ .Values.authProxy.image.tag }}"
+          imagePullPolicy: {{ .Values.authProxy.image.pullPolicy }}
+          volumeMounts:
+            - mountPath: /etc/oauth2_proxy
+              name: {{ auth-proxy-config }}
```
andrewazores, line 200: I don't think this needs to be wrapped in curly braces - it's just a reference to the literal `name: auth-proxy-config` lower down in this same file. The curly braces are for variable interpolation.

```
$ helm install cryostat --set pvc.enabled=true --set core.route.enabled=true ./charts/cryostat
Error: INSTALLATION FAILED: parse error at (cryostat/templates/deployment.yaml:200): bad character U+002D '-'
```

charts/cryostat/templates/deployment.yaml
```diff
@@ -190,6 +190,14 @@ spec:
           securityContext:
             {{- toYaml .Values.storage.securityContext | nindent 12 }}
         {{- if not .Values.minimal }}
+        - name: {{ printf "%s-%s" .Chart.Name "auth-proxy" }}
+          securityContext:
+            {{- toYaml .Values.authProxy.securityContext | nindent 12 }}
+          image: "{{ .Values.authProxy.image.repository }}:{{ .Values.authProxy.image.tag }}"
```
andrewazores, line 196: It doesn't look like `authProxy.image.repository` and `authProxy.image.tag` are defined in `values.yaml`? Same with the `.securityContext` and `image.pullPolicy` etc.

charts/cryostat/values.yaml
```diff
@@ -198,6 +198,16 @@ authentication:
     ## @param authentication.basicAuth.filename Key within Secret containing the properties file. The properties file should contain one user per line, with the syntax "user=passHex", where "user" is the username and "passHex" is the SHA-256 hash of the desired password **(Required if basicAuth is enabled)**
     filename: ""
 
+oauth2Proxy:
+  ## @param authentication.oauth2Proxy.enabled Whether Cryostat should use OAuth2 Proxy for authentication. When true, Cryostat will authenticate users via OAuth2 Proxy.
+  enabled: true
+  ## @param authentication.oauth2Proxy.htpasswdFile Path to the htpasswd file for additional user authentication
+  htpasswdFile: "templates/auth_proxy_htpasswd"
+  ## @param authentication.oauth2Proxy.cookieSecret Secret key for encrypting OAuth2 Proxy cookies
+  cookieSecret: __24_BYTE_COOKIE_SECRET_
```
andrewazores, line 207: This should allow the user to specify an override if they have some particular need, but if they do not then it should not default to a hardcoded value. Then it isn't a "secret" anymore :-) this should use a generated value like other secrets within `_helpers.tpl`.

.github/workflows/helm-release.yml
aali309: ooh wait, I dint make this change!

charts/cryostat/values.yaml
```diff
@@ -198,6 +198,16 @@ authentication:
     ## @param authentication.basicAuth.filename Key within Secret containing the properties file. The properties file should contain one user per line, with the syntax "user=passHex", where "user" is the username and "passHex" is the SHA-256 hash of the desired password **(Required if basicAuth is enabled)**
     filename: ""
 
+oauth2Proxy:
+  ## @param authentication.oauth2Proxy.enabled Whether Cryostat should use OAuth2 Proxy for authentication. When true, Cryostat will authenticate users via OAuth2 Proxy.
+  enabled: true
+  ## @param authentication.oauth2Proxy.htpasswdFile Path to the htpasswd file for additional user authentication
+  htpasswdFile: "templates/auth_proxy_htpasswd"
```
andrewazores, line 205: I think this overlaps with the existing `.Values.authentication.basicAuth` configuration. We should keep that configuration, which allows the user to specify a secret that contains the basic auth credentials, and if they do that then we should mount that secret into the oauth2-proxy at this hardcoded location.

charts/cryostat/values.yaml
```diff
@@ -198,6 +198,16 @@ authentication:
     ## @param authentication.basicAuth.filename Key within Secret containing the properties file. The properties file should contain one user per line, with the syntax "user=passHex", where "user" is the username and "passHex" is the SHA-256 hash of the desired password **(Required if basicAuth is enabled)**
     filename: ""
 
+oauth2Proxy:
+  ## @param authentication.oauth2Proxy.enabled Whether Cryostat should use OAuth2 Proxy for authentication. When true, Cryostat will authenticate users via OAuth2 Proxy.
+  enabled: true
```
andrewazores, line 203: This is the same as `.Values.authentication.basicAuth.enabled` - let's keep that one and remove this.

.github/workflows/helm-release.yml
andrewazores: Not sure how it snuck in here then - either way, please remove it from this PR.

charts/cryostat/templates/deployment.yaml
```diff
@@ -190,6 +190,23 @@ spec:
           securityContext:
             {{- toYaml .Values.storage.securityContext | nindent 12 }}
         {{- if not .Values.minimal }}
+        - name: {{ printf "%s-%s" .Chart.Name "oauth2proxy" }}
+          securityContext:
+            {{- toYaml .Values.oauth2Proxy.securityContext | nindent 12 }}
+          image: "{{ .Values.oauth2Proxy.image.repository }}:{{ .Values.oauth2Proxy.image.tag }}"
+          imagePullPolicy: {{ .Values.oauth2Proxy.image.pullPolicy }}
+          env:
+            - name: CRYOSTAT_OAUTH2PROXY_URL
```
andrewazores, line 199: I don't think the name of this environment variable should have been changed. The oauth2-proxy container won't know how to interpret a variable named like this.

charts/cryostat/values.yaml
```diff
@@ -198,6 +198,29 @@ authentication:
...
+  ## @param authentication.oauth2Proxy.htpasswdFile Path to the htpasswd file for additional user authentication
+  htpasswdFile: "templates/auth_proxy_htpasswd"
+  ## @param authentication.oauth2Proxy.cookieSecret Secret key for encrypting OAuth2 Proxy cookies
+  cookieSecret: { { include "cryostat.defaultCookieSecret" . } }
+  image:
+    ## @param authentication.oauth2Proxy.image.repository Repository for the OAuth2 Proxy container image
+    repository: "quay.io/cryostat/oauth2-proxy"
```
andrewazores, line 210: We don't host this image in our own repository namespace, it should be: `quay.io/oauth2-proxy/oauth2-proxy:latest`

.github/workflows/helm-release.yml
aali309: It came from Thuan's commit

charts/cryostat/templates/deployment.yaml
```diff
@@ -190,6 +190,23 @@ spec:
           securityContext:
             {{- toYaml .Values.storage.securityContext | nindent 12 }}
         {{- if not .Values.minimal }}
+        - name: {{ printf "%s-%s" .Chart.Name "oauth2proxy" }}
+          securityContext:
+            {{- toYaml .Values.oauth2Proxy.securityContext | nindent 12 }}
+          image: "{{ .Values.oauth2Proxy.image.repository }}:{{ .Values.oauth2Proxy.image.tag }}"
+          imagePullPolicy: {{ .Values.oauth2Proxy.image.pullPolicy }}
+          env:
+            - name: redirectURL
```
andrewazores, line 199: I think this environment variable must be named `OAUTH2_PROXY_REDIRECT_URL`...

charts/cryostat/templates/deployment.yaml
```diff
@@ -190,6 +190,23 @@ spec:
...
+              value: "http://localhost:8181/oauth2/callback"
```
andrewazores, line 200: Should this be port 8181? I would expect it to be 8080 as the default for this container, since 8181 is Cryostat's HTTP port.

aali309: right..

andrewazores:
Rebase on top of upstream/cryostat3 again please - I just updated that to be up to date with `main`.

charts/cryostat/values.yaml
andrewazores:
```
$ helm install cryostat ./charts/cryostat
Error: INSTALLATION FAILED: cannot load values.yaml: error converting YAML to JSON: yaml: invalid map key: map[interface {}]interface {}{"include \"cryostat.defaultCookieSecret\" .":interface {}(nil)}
```

I think this can just be removed from `values.yaml`. If you look at how `cryostat.objectStorageSecretKey` and `-storage-secret-key` are used it should be a good example to follow for how this can be put together. `cryostat.databaseEncryptionKey` and `db-encryption-key` as well.

And please do try running `helm install` yourself after making changes so you can validate whether the changes are installable or produce the expected deployment outcome.

.github/workflows/helm-release.yml
```diff
@@ -39,7 +39,7 @@ jobs:
             -f target_commitish="${GITHUB_REF_NAME}" \
             -f previous_tag_name="${{ steps.previous-version.outputs.tag }}" | jq -r .body > charts/cryostat/release-notes.md
       - name: Run chart-releaser
-        uses: helm/chart-releaser-action@v1.6.0
+        uses: helm/chart-releaser-action@v1.5.0
```
andrewazores, line 42: This still needs to be removed from the PR, please

charts/cryostat/templates/auth2_proxy.yaml
```diff
@@ -0,0 +1,6 @@
+apiVersion: v1
+kind: Secret
+metadata:
+  name: {{ .Release.Name }}-auth-proxy-secret
+data:
+  users.auth_proxy_htpasswd: {{ .Files.Get "templates/auth_proxy_htpasswd" | b64enc | quote }}
```
andrewazores, line 6: https://helm.sh/docs/chart_template_guide/accessing_files/

> Some files cannot be accessed through the .Files object, usually for security reasons.
>    Files in templates/ cannot be accessed.

andrewazores: Also,

```yaml
data:
  users.auth_proxy_htpasswd: something
```

is an unusual looking format for YAML - normally an object key would just be `users` and `auth_proxy_htpasswd` would be a key of a sub-object, or something like that. Maybe this also needs to be changed?

aali309: > https://helm.sh/docs/chart_template_guide/accessing_files/
> 
> > Some files cannot be accessed through the .Files object, usually for security reasons.
> > Files in templates/ cannot be accessed.

Thank you for this!

aali309: > Also,
> 
> ```yaml
> data:
>   users.auth_proxy_htpasswd: something
> ```
> 
> is an unusual looking format for YAML - normally an object key would just be `users` and `auth_proxy_htpasswd` would be a key of a sub-object, or something like that. Maybe this also needs to be changed?

It works the same with both `users.fileName: path` or just `fileName:path` the issue was accessing the file inside templates.

andrewazores:
This results in a deployment, but a failing one:

image

image

`$oc describe pod`:
```
  cryostat-oauth2proxy:
    Container ID:   cri-o://f48983f0b3e61fe2205618c626bb18cef5d50500237b2101fa9d70d1f81759ec
    Image:          quay.io/oauth2-proxy/oauth2-proxy:latest
    Image ID:       quay.io/oauth2-proxy/oauth2-proxy@sha256:3c6012940d3cd71ecc67fd7783916df6d27972b520209af36e9567b9bdbc0c79
    Port:           <none>
    Host Port:      <none>
    State:          Waiting
      Reason:       CrashLoopBackOff
    Last State:     Terminated
      Reason:       Error
      Exit Code:    1
      Started:      Thu, 25 Jan 2024 15:24:04 -0500
      Finished:     Thu, 25 Jan 2024 15:24:04 -0500
    Ready:          False
    Restart Count:  3
    Environment:
      OAUTH2_PROXY_REDIRECT_URL:  http://localhost:8080/oauth2/callback
    Mounts:
      /etc/oauth2_proxy from auth-proxy-config (ro)
      /var/run/secrets/kubernetes.io/serviceaccount from kube-api-access-ptj2t (ro)
```

`$ oc logs -f -c cryostat-oauth2proxy pod/cryostat-7776fbc9fc-dvnpl`
```
[2024/01/25 20:24:46] [main.go:54] invalid configuration:
  missing setting: cookie-secret
  provider missing setting: client-id
  missing setting: client-secret or client-secret-file
  missing setting for email validation: email-domain or authenticated-emails-file required.
      use email-domain=* to authorize all email addresses
```

andrewazores:
`cookie-secret` can be provided by the environment variable `OAUTH2_PROXY_COOKIE_SECRET`, where you would pass in the secret you generated.

`client-id` should be in the config file.

https://github.com/cryostatio/cryostat3/blob/20f05eacd12e6256b19316d21411714f7eea80db/smoketest/compose/auth_proxy.yml#L41

https://github.com/cryostatio/cryostat3/blob/main/smoketest/compose/auth_proxy_alpha_config.yaml

charts/cryostat/templates/deployment.yaml
```diff
@@ -194,6 +194,35 @@ spec:
...
+            - name: OAUTH2_PROXY_PROVIDER
+              value: "google"
+            - name: OAUTH2_PROXY_CLIENT_ID
+              value: "CLIENT_ID"
+            - name: OAUTH2_PROXY_CLIENT_SECRET
+              value: "CLIENT_SECRET"
+            - name: OAUTH2_PROXY_EMAIL_DOMAINS
+              value: "*"
+            - name: UPSTREAM_CONFIG
+              value: {{ include "cryostat.base64Encode" "proxyRawPath: true" }}
```
andrewazores, line 216: This doesn't look quite right... I think the alpha config must be passed as a path on the container start command, like this:

https://github.com/cryostatio/cryostat3/blob/328402ef8989f9aff9aec414cba265fdaae9f36f/smoketest/compose/auth_proxy.yml#L32

and the config should be a yaml file, not a base64-encoded string, like this one:

https://github.com/cryostatio/cryostat3/blob/main/smoketest/compose/auth_proxy_alpha_config.yaml

ebaron: I think we'll need to account for this alpha config somehow in downstream builds. It didn't look like the OpenShift oauth-proxy supports this.

charts/cryostat/templates/auth_proxy_alpha_config.yaml
```diff
@@ -0,0 +1,29 @@
...
+          rewriteTarget: /$1
+          uri: http://s3:3000
+          passHostHeader: false
+          proxyWebSockets: false
+    providers:
+      - id: dummy
+        name: Unused - Sign In Below
+        clientId: CLIENT_ID
+        clientSecret: CLIENT_SECRET
+        provider: google
```
ebaron, line 29: Maybe we should take the same approach here that you used for the htpasswd file. Extract out this upstreamConfig YAML file into its own file and then use the `Files` template function with the `.oauth2Proxy.alphaConfig.confgFile` value, defaulting to the location of that YAML file.

aali309: I  tried this, ended up not reading the file or file not found.

charts/cryostat/templates/deployment.yaml
andrewazores: That's correct, unfortunately. We need it for oauth2-proxy because that's how you specify the behaviour to not re-encode path segments, and they only expose that config property through this YAML alpha config. On openshift-oauth-proxy the bugged re-encoding is always done, but it looks like with the upcoming Go version update it will not do it anymore and the bug is fixed.

I'm not sure how both projects will end up handling it since this is a behavioural change in the underlying Go standard library. oauth2-proxy doesn't use that anymore and instead uses a library dependency that allows them to have this configurable switch because they were afraid of changing behaviour and breaking compatibility. I don't know what openshift-oauth-proxy will end up doing or if they have the same concern (ie applications built to expect the proxy's bugged behaviour and work around it).

andrewazores: My idea is that we will add some configuration switch for users to enable OpenShift integration. Turning that on is what will instruct the helm chart to deploy openshift-oauth-proxy instead of oauth2-proxy, so this will mean that not only the deployed image changes, but these specifics of how the containers are configured will change. That will also be an upstream feature, but deploying with openshift-oauth-proxy will always result in a broken deployment until they fix the path segment re-encoding bug anyway.

andrewazores:
```
I0207 19:13:09.695531 volume.go:370 Start Seaweed volume server 30GB 3.62  at 0.0.0.0:8080
F0207 19:13:09.695587 volume.go:373 Volume server listener error:listen tcp 0.0.0.0:8080: bind: address already in use
```

When I try to install this now, I get this message in the `cryostat-storage` container logs and the Pod does not become Ready because the storage container fails liveness checks (due to the above error).

I think this is coming from the recent changes to also put more of the other containers behind the same proxy instance. Now that they are all part of one pod there is a port conflict where both `-storage` and something elseare both trying to bind to the same container port.

We have an open issue to make the `-storage` port numbers configurable: https://github.com/cryostatio/cryostat-storage/issues/2 . So either we need to tackle that, or figure out what else is trying to bind on 8080 and see if that number can be changed, or figure out what else is trying to bind on port 8080 and move that outside of the Pod.

aali309:
binding on 4180 i.e ` BindAddress: http://0.0.0.0:4180` still storage fails.

```
Events:
  Type     Reason     Age                   From               Message
  ----     ------     ----                  ----               -------
  Normal   Scheduled  22m                   default-scheduler  Successfully assigned default/cryostat-848bfc65bb-hthtp to kind-control-plane
  Normal   Pulled     20m                   kubelet            Successfully pulled image "quay.io/cryostat/cryostat:3.0.0-snapshot" in 1m28.774941707s
  Normal   Created    20m                   kubelet            Created container cryostat
  Normal   Started    20m                   kubelet            Started container cryostat
  Normal   Pulling    20m                   kubelet            Pulling image "quay.io/cryostat/cryostat-db:latest"
  Normal   Pulled     20m                   kubelet            Successfully pulled image "quay.io/cryostat/cryostat-db:latest" in 45.180998903s
  Normal   Created    19m                   kubelet            Created container cryostat-db
  Normal   Started    19m                   kubelet            Started container cryostat-db
  Normal   Pulling    19m                   kubelet            Pulling image "quay.io/oauth2-proxy/oauth2-proxy:latest"
  Normal   Pulled     19m                   kubelet            Successfully pulled image "quay.io/cryostat/cryostat-storage:latest" in 5.493475043s
  Warning  Failed     19m                   kubelet            Error: container has runAsNonRoot and image will run as root (pod: "cryostat-848bfc65bb-hthtp_default(6eaa0b35-b57b-45ab-bb26-5e84242186e7)", container: cryostat-storage)
  Normal   Started    19m                   kubelet            Started container cryostat-oauth2proxy
  Normal   Pulling    19m                   kubelet            Pulling image "quay.io/cryostat/cryostat-grafana-dashboard:latest"
  Normal   Pulled     19m                   kubelet            Successfully pulled image "quay.io/oauth2-proxy/oauth2-proxy:latest" in 2.399092856s
  Normal   Created    19m                   kubelet            Created container cryostat-oauth2proxy
  Normal   Pulled     19m                   kubelet            Successfully pulled image "quay.io/cryostat/cryostat-grafana-dashboard:latest" in 26.645259539s
  Normal   Created    19m                   kubelet            Created container cryostat-grafana
  Normal   Started    19m                   kubelet            Started container cryostat-grafana
  Normal   Pulling    19m                   kubelet            Pulling image "quay.io/cryostat/jfr-datasource:latest"
  Normal   Pulled     19m                   kubelet            Successfully pulled image "quay.io/cryostat/jfr-datasource:latest" in 4.653059035s
  Normal   Created    19m                   kubelet            Created container cryostat-jfr-datasource
  Normal   Started    19m                   kubelet            Started container cryostat-jfr-datasource
  Normal   Pulling    19m (x2 over 22m)     kubelet            Pulling image "quay.io/cryostat/cryostat:3.0.0-snapshot"
  Normal   Pulled     19m                   kubelet            Successfully pulled image "quay.io/cryostat/cryostat:3.0.0-snapshot" in 605.139021ms
  Normal   Pulled     7m14s (x57 over 19m)  kubelet            (combined from similar events): Successfully pulled image "quay.io/cryostat/cryostat-storage:latest" in 431.580772ms
  Normal   Pulling    2m4s (x81 over 19m)   kubelet            Pulling image "quay.io/cryostat/cryostat-storage:latest"
```

andrewazores:
Could you post the container's logs or push your changes so I can take a look at it too?

andrewazores:
I just tried running it explicitly as a non-root user in `kind` and that seemed to be OK. I'm starting my `crc` instance now to test it there as well. If this works then the proper fix will be to patch the `cryostat-storage` Dockerfile to specify a non-root default user, I think, but I'll have to do a little more reading and refreshing my memory on how this is supposed to work.

andrewazores:
Not working in `crc`. OpenShift's different way of doing things is leading to permissions issues for `-storage` still, as well as for `-db`. Not sure what changed there, I had that working before.

andrewazores:
Ah, right - it's broken if I just use the `default` namespace in `crc`. If I drop that last commit I made, then first do `oc new-project cryostat3` (or any other name) before `helm install` inside that new project then I don't run into these permissions issues.

```bash
$ git pull
$ git revert --no-commit HEAD
$ helm uninstall cryostat
$ oc new-project cryostat3
$ helm install cryostat --set pvc.enabled=true --set core.route.enabled=true ./charts/cryostat
```

Can you try that and see if it works for you as well?

andrewazores:
I'll revert that commit on this branch so that you don't have to do that step manually, since the change isn't really useful (at this time anyway).

aali309:
For me just changing the binding port works well on OpenShift. 

`oc new-project test`
`helm install cryostat ./charts/cryostat --set platform=openshift`

image

image

andrewazores:
Looks good - yep, the port number still needed to be changed as well so that the storage and authproxy don't conflict on 8080.

andrewazores:
Okay, so now that we've figured out how to get it deployable and running again... back to reviewing the important changes from this PR.

charts/cryostat/auth_proxy_htpasswd
```diff
@@ -0,0 +1 @@
+user:$2y$05$.p1/68oBWKX1FhAtZOArYOoNoMqsB4yuUNXGOeSATHPq3geKqEaZS
```
andrewazores, line 1: It doesn't seem like the authproxy is actually using this `htpasswd` file. When I `helm install` now I'm still able to just open the `route` and go straight into the Cryostat Web UI without having to authenticate to the proxy.

I suspect the problem is that the `service` and/or `route` are still being defined so that traffic is flowing directly into the Cryostat container, whereas it should now be flowing through the authproxy conntainer (which will then proxy to Cryostat and the other containers).

andrewazores: Yep - I think `service.yaml` at least needs adjusting so that it points to `8080` (authproxy) instead of `8181` (cryostat3).

andrewazores: Sorry - that should be `4180` now with the port binding change, not `8080`.

charts/cryostat/templates/deployment.yaml
```diff
@@ -194,6 +194,31 @@ spec:
...
+          args:
+            - "--alpha-config=/etc/oauth2_proxy/alpha_config/alpha_config.yaml"
+          imagePullPolicy: {{ .Values.oauth2Proxy.image.pullPolicy }}
+          env:
+            - name: OAUTH2_PROXY_REDIRECT_URL
+              value: "http://localhost:8080/oauth2/callback"
+            - name: OAUTH2_PROXY_COOKIE_SECRET
+              value: {{ include "cryostat.cookieSecret" . }}
+            - name: OAUTH2_PROXY_EMAIL_DOMAINS
+              value: "*"
```
andrewazores, line 210: I think this also needs environment variables to tell it to use the `htpasswd` and what permissions to assign to users it finds there, like this:

https://github.com/cryostatio/cryostat3/blob/a85f4900250969578a93526fd4bbe61e3fbe4ef2/smoketest/compose/auth_proxy.yml#L41

charts/cryostat/templates/alpha_config.yaml
```diff
@@ -0,0 +1,29 @@
...
+data:
+  alpha_config.yaml: |-
+    server:
+      BindAddress: http://0.0.0.0:4180
+    upstreamConfig:
+      proxyRawPath: true
+      upstreams:
+        - id: cryostat
+          path: /
+          uri: http://cryostat:8181
```
andrewazores, line 14: I think the hostnames here (`cryostat`, `grafana`, `s3`) in the URLs should actually just be `localhost`. They are distinct hostnames in the smoketest version because there each of these containers are run separately and are addressable by their container names. In this case we are deploying them all as containers within one Pod, so I think just using `localhost` is what will work most simply.

andrewazores:
I think with my latest changes this is more or less working. When you open the `route` URL you hit the oauth2-proxy login page with the htpasswd login form, just like in the cryostat3 smoketest. When you use `user:pass` to log in then you see the Cryostat UI and can interact with it as usual.

Next steps for you to tackle, please:

1. Fix up the 2 `auth2_proxy.yaml`. In my commit it's just using a hardcoded value representing `user:pass`. In your previous commits it was trying to use a hardcoded file to use the same values. We don't want to hardcode the values at all - these must be configured and supplied by the user. This means that the user must create their own Secret with their own contents (ie `user:pass` or something else, with the correct encoding etc already). Then the user passes this via `values.yaml` `.authentication.basicAuth` configuration.
2. If `values.yaml` `.authentication.basicAuth.enabled = false` then set the env var `OAUTH2_PROXY_SKIP_AUTH_ROUTES=.*` on the authproxy container. This way the auth proxy is always in place, but if the user does not enable authentication, then the proxy just passes requests through without requiring the user to authenticate.
3. `values.yaml` section for `oauth2Proxy` should not have the `htpasswdFile` property - I think this is doing the same thing that `.authentication.basicAuth.secretName`/`.authentication.basicAuth.filename` is doing. Also, `.oauth2Proxy.alphaConfig` looks like it should probably be removed and always hardcoded - I can't see that the user would need to be able to change these configurations at this level of granularity.

----

For separate PR(s):
4. env var `STORAGE_EXT_URL` needs to be set on the Cryostat container.
5. Later when we add support for deploying `openshift-oauth-proxy` in place of `oauth2-proxy`, the `values.yaml` `.oauth2Proxy.enabled` configuration should also be replaced with something else like `.core.openshift-integration.enabled`. This would switch on both `openshift-oauth-proxy` and I think should also replace `.core.route.enabled`.

charts/cryostat/templates/auth2_proxy.yaml
```diff
@@ -3,6 +3,4 @@ kind: Secret
 metadata:
   name: {{ .Release.Name }}-auth-proxy-secret
 data:
-  # auth_proxy_htpasswd: {{ .Files.Get ".oauth2proxy.htpasswdFile" | b64enc | quote }}
-  # FIXME this is a hack, the user definition should be pulled out of values.yaml: .authentication.basicAuth
-  auth_proxy_htpasswd: {{ "user:$2y$05$.p1/68oBWKX1FhAtZOArYOoNoMqsB4yuUNXGOeSATHPq3geKqEaZS" | b64enc | quote }}
+    auth_proxy_htpasswd: {{ .Values.authentication.basicAuth.secretName | b64enc | quote }}
```
andrewazores, line 6: This means that the helm chart will create a new Secret named ex. `cryostat-auth-proxy-secret`, containing the data `auth_proxy_htpasswd: another-secret-name`. That doesn't seem quite right - we don't want/need to have this second level of indirection with a Secret pointing to another Secret.

Instead, the user's provided Secret should be mounted as a volume, using the `basicAuth.secretName` and `basicAuth.filename` they provide.

https://kubernetes.io/docs/concepts/configuration/secret/#using-secrets-as-files-from-a-pod

In `deployment.yaml` we already have parts of that set up:

```yaml
      {{- if .Values.authentication.basicAuth.enabled }}
      - name: basic-auth-secret
        secret:
          secretName: {{ .Values.authentication.basicAuth.secretName }}
      {{- end }}
```

```yaml
            {{- if .Values.authentication.basicAuth.enabled }}
            - name: basic-auth-secret
              mountPath: /etc/oauth2_proxy/basic-auth-secret
              readOnly: true
            {{- end }}
```

I think `auth2_proxy.yaml` should be removed, and this yaml above from `deployment.yaml` just needs to be adjusted to use `basicAuth.secretName` and `basicAuth.filename` to mount the user's provided Secret into the path of `/etc/oauth2_proxy/auth_proxy_htpasswd`.

Also, you'll have to put this back if `basicAuth.enabled`:

```diff
-            - name: OAUTH2_PROXY_HTPASSWD_FILE
-              value: /etc/oauth2_proxy/auth_proxy_htpasswd
```

charts/cryostat/templates/deployment.yaml
```diff
@@ -208,10 +208,12 @@ spec:
               value: {{ include "cryostat.cookieSecret" . }}
             - name: OAUTH2_PROXY_EMAIL_DOMAINS
               value: "*"
-            - name: OAUTH2_PROXY_HTPASSWD_FILE
-              value: /etc/oauth2_proxy/auth_proxy_htpasswd
```
andrewazores: In short from above, don't remove this piece. Wrap it in `{{ if .Values.authentication.basicAuth.enabled }}`, so that we set this environment variable telling the proxy to load the htpasswd file only when the user has enabled basic auth.

andrewazores:
For example, I created this secret manually:

```yaml
apiVersion: v1
data:
  htpasswd: dXNlcjokMnkkMDUkLnAxLzY4b0JXS1gxRmhBdFpPQXJZT29Ob01xc0I0eXVVTlhHT2VTQVRIUHEzZ2VLcUVhWlM=
kind: Secret
metadata:
  creationTimestamp: "2024-02-09T17:07:49Z"
  name: basicauth
  namespace: cryostat3
  resourceVersion: "75991"
  uid: 16eab9c1-8dd3-4b64-b544-cbb4d52cccba
type: Opaque
```

This is still just `user:pass` like before.

and I applied this patch on top of this branch:

```diff
diff --git a/charts/cryostat/templates/auth2_proxy.yaml b/charts/cryostat/templates/auth2_proxy.yaml
deleted file mode 100644
index e49b4b1..0000000
--- a/charts/cryostat/templates/auth2_proxy.yaml
+++ /dev/null
@@ -1,6 +0,0 @@
-apiVersion: v1
-kind: Secret
-metadata:
-  name: {{ .Release.Name }}-auth-proxy-secret
-data:
-    auth_proxy_htpasswd: {{ .Values.authentication.basicAuth.secretName | b64enc | quote }}
diff --git a/charts/cryostat/templates/deployment.yaml b/charts/cryostat/templates/deployment.yaml
index 88536bd..d420158 100644
--- a/charts/cryostat/templates/deployment.yaml
+++ b/charts/cryostat/templates/deployment.yaml
@@ -208,21 +208,22 @@ spec:
               value: {{ include "cryostat.cookieSecret" . }}
             - name: OAUTH2_PROXY_EMAIL_DOMAINS
               value: "*"
+            {{- if .Values.authentication.basicAuth.enabled }}
             - name: OAUTH2_PROXY_HTPASSWD_USER_GROUP
               value: write
+            - name: OAUTH2_PROXY_HTPASSWD_FILE
+              value: /etc/oauth2_proxy/basicauth/{{ .Values.authentication.basicAuth.filename }}
+            {{- end }}
             {{- if not .Values.authentication.basicAuth.enabled }}
             - name: OAUTH2_PROXY_SKIP_AUTH_ROUTES
               value: ".*"
-          {{- end }}
+            {{- end }}
           volumeMounts:
-            - name: auth-proxy-config
-              mountPath: /etc/oauth2_proxy
-              readOnly: true
             - name: alpha-config
               mountPath: /etc/oauth2_proxy/alpha_config
             {{- if .Values.authentication.basicAuth.enabled }}
-            - name: basic-auth-secret
-              mountPath: /etc/oauth2_proxy/basic-auth-secret
+            - name: {{ .Release.Name }}-htpasswd
+              mountPath: /etc/oauth2_proxy/basicauth
               readOnly: true
             {{- end }}
         - name: {{ printf "%s-%s" .Chart.Name "grafana" }}
@@ -291,11 +292,8 @@ spec:
       - name: alpha-config
         configMap:
           name: {{ .Release.Name }}-alpha-config
-      - name: auth-proxy-config
-        secret:
-          secretName: {{ .Release.Name }}-auth-proxy-secret
       {{- if .Values.authentication.basicAuth.enabled }}
-      - name: basic-auth-secret
+      - name: {{ .Release.Name }}-htpasswd
         secret:
           secretName: {{ .Values.authentication.basicAuth.secretName }}
       {{- end }}
```

With this, I can do:

```bash
$ helm install cryostat --set pvc.enabled=true --set core.route.enabled=true --set authentication.basicAuth.enabled=false --set authentication.basicAuth.secretName=basicauth --set authentication.basicAuth.filename=htpasswd ./charts/cryostat
```

and get a deployment that has an authproxy set up but which passes all requests without authentication. Simply changing that to `authentication.basicAuth.enabled=true`, like this:

```bash
$ helm install cryostat --set pvc.enabled=true --set core.route.enabled=true --set authentication.basicAuth.enabled=true --set authentication.basicAuth.secretName=basicauth --set authentication.basicAuth.filename=htpasswd ./charts/cryostat
```

gets me a deployment that uses the Secret I created to perform basic auth on the proxy.

aali309:
> For example, I created this secret manually:
> 
> ```yaml
> apiVersion: v1
> data:
>   htpasswd: dXNlcjokMnkkMDUkLnAxLzY4b0JXS1gxRmhBdFpPQXJZT29Ob01xc0I0eXVVTlhHT2VTQVRIUHEzZ2VLcUVhWlM=
> kind: Secret
> metadata:
>   creationTimestamp: "2024-02-09T17:07:49Z"
>   name: basicauth
>   namespace: cryostat3
>   resourceVersion: "75991"
>   uid: 16eab9c1-8dd3-4b64-b544-cbb4d52cccba
> type: Opaque
> ```
> 
> This is still just `user:pass` like before.
> 
> and I applied this patch on top of this branch:
> 
> ```diff
> diff --git a/charts/cryostat/templates/auth2_proxy.yaml b/charts/cryostat/templates/auth2_proxy.yaml
> deleted file mode 100644
> index e49b4b1..0000000
> --- a/charts/cryostat/templates/auth2_proxy.yaml
> +++ /dev/null
> @@ -1,6 +0,0 @@
> -apiVersion: v1
> -kind: Secret
> -metadata:
> -  name: {{ .Release.Name }}-auth-proxy-secret
> -data:
> -    auth_proxy_htpasswd: {{ .Values.authentication.basicAuth.secretName | b64enc | quote }}
> diff --git a/charts/cryostat/templates/deployment.yaml b/charts/cryostat/templates/deployment.yaml
> index 88536bd..d420158 100644
> --- a/charts/cryostat/templates/deployment.yaml
> +++ b/charts/cryostat/templates/deployment.yaml
> @@ -208,21 +208,22 @@ spec:
>                value: {{ include "cryostat.cookieSecret" . }}
>              - name: OAUTH2_PROXY_EMAIL_DOMAINS
>                value: "*"
> +            {{- if .Values.authentication.basicAuth.enabled }}
>              - name: OAUTH2_PROXY_HTPASSWD_USER_GROUP
>                value: write
> +            - name: OAUTH2_PROXY_HTPASSWD_FILE
> +              value: /etc/oauth2_proxy/basicauth/{{ .Values.authentication.basicAuth.filename }}
> +            {{- end }}
>              {{- if not .Values.authentication.basicAuth.enabled }}
>              - name: OAUTH2_PROXY_SKIP_AUTH_ROUTES
>                value: ".*"
> -          {{- end }}
> +            {{- end }}
>            volumeMounts:
> -            - name: auth-proxy-config
> -              mountPath: /etc/oauth2_proxy
> -              readOnly: true
>              - name: alpha-config
>                mountPath: /etc/oauth2_proxy/alpha_config
>              {{- if .Values.authentication.basicAuth.enabled }}
> -            - name: basic-auth-secret
> -              mountPath: /etc/oauth2_proxy/basic-auth-secret
> +            - name: {{ .Release.Name }}-htpasswd
> +              mountPath: /etc/oauth2_proxy/basicauth
>                readOnly: true
>              {{- end }}
>          - name: {{ printf "%s-%s" .Chart.Name "grafana" }}
> @@ -291,11 +292,8 @@ spec:
>        - name: alpha-config
>          configMap:
>            name: {{ .Release.Name }}-alpha-config
> -      - name: auth-proxy-config
> -        secret:
> -          secretName: {{ .Release.Name }}-auth-proxy-secret
>        {{- if .Values.authentication.basicAuth.enabled }}
> -      - name: basic-auth-secret
> +      - name: {{ .Release.Name }}-htpasswd
>          secret:
>            secretName: {{ .Values.authentication.basicAuth.secretName }}
>        {{- end }}
> ```
> 
> With this, I can do:
> 
> ```shell
> $ helm install cryostat --set pvc.enabled=true --set core.route.enabled=true --set authentication.basicAuth.enabled=false --set authentication.basicAuth.secretName=basicauth --set authentication.basicAuth.filename=htpasswd ./charts/cryostat
> ```
> 
> and get a deployment that has an authproxy set up but which passes all requests without authentication. Simply changing that to `authentication.basicAuth.enabled=true`, like this:
> 
> ```shell
> $ helm install cryostat --set pvc.enabled=true --set core.route.enabled=true --set authentication.basicAuth.enabled=true --set authentication.basicAuth.secretName=basicauth --set authentication.basicAuth.filename=htpasswd ./charts/cryostat
> ```
> 
> gets me a deployment that uses the Secret I created to perform basic auth on the proxy.

yes I have done something similar... 
i get this error  from storage as before: 
```Warning  Failed     70s                kubelet            Error: container has runAsNonRoot and image will run as root (pod: "cryostat-688d4f9d5c-582w5_default(f83c8023-a0df-4385-a9a7-9abb78b24910)", container: cryostat-storage)
```
but runs well on OpenShift

andrewazores:
Are you in the `default` namespace/project again? I don't think any of the changes made here should have any effect on whether the root user is used.

aali309:
Looks good now on my side as well now too. 
when basicAuth is enabled and no secret defined 
`❯ helm install cryostat ./charts/cryostat --set platform=openshift

Error: INSTALLATION FAILED: Deployment.apps "cryostat" is invalid: [spec.template.spec.volumes[2].secret.secretName: Required value, spec.template.spec.containers[3].volumeMounts[1].name: Not found: "cryostat-htpasswd"]
`
which makes sense according to what you were telling me earlier.
when disabled no authentication is required.

andrewazores (approved):
I think this looks good now, but let's wait for @ebaron to be back and give it a look-over too.

ebaron (commented):
This looks mostly good, but I think there's still the odd bug. When I tested using the default `ClusterIP` service type, I didn't get directed to the OAuth login page, but to Cryostat itself.

Here's how I tested:
1. `$ htpasswd -nbB foo bar > /tmp/htpasswd`
2. `$ oc create secret generic htpasswd --from-file=htpasswd=/tmp/htpasswd`
3. `$ helm install cryostat ./charts/cryostat/ --set authentication.basicAuth.enabled=true,authentication.basicAuth.secretName=htpasswd,authentication.basicAuth.filename=htpasswd`
 
This is the output of the NOTES:
1. Tell Cryostat how to serve external traffic:
  ```
  export POD_NAME=$(kubectl get pods -n helm-test -l "app.kubernetes.io/name=cryostat,app.kubernetes.io/instance=cryostat"  --sort-by=.metadata.creationTimestamp -o jsonpath="{.items[-1:].metadata.name}")
  export CONTAINER_PORT=$(kubectl get pod -n helm-test $POD_NAME -o jsonpath="{.spec.containers[0].ports[0].containerPort}")
  export GRAFANA_CONTAINER_PORT=$(kubectl get pod -n helm-test $POD_NAME -o jsonpath="{.spec.containers[1].ports[0].containerPort}")
  kubectl -n helm-test set env deploy --containers=cryostat cryostat CRYOSTAT_WEB_HOST=127.0.0.1 CRYOSTAT_EXT_WEB_PORT=8080 GRAFANA_DASHBOARD_URL=http://127.0.0.1:$GRAFANA_CONTAINER_PORT GRAFANA_DASHBOARD_EXT_URL=http://127.0.0.1:8081
  ```

2. Forward local ports to the application's pod:
  ```
  export POD_NAME=$(kubectl get pods -n helm-test -l "app.kubernetes.io/name=cryostat,app.kubernetes.io/instance=cryostat"  --sort-by=.metadata.creationTimestamp -o jsonpath="{.items[-1:].metadata.name}")
  kubectl -n helm-test wait --for=condition=available --timeout=60s deploy/cryostat
  kubectl -n helm-test port-forward $POD_NAME 8080:$CONTAINER_PORT 8081:$GRAFANA_CONTAINER_PORT
  ```

3. Visit the Cryostat application at: 
  ```
  http://127.0.0.1:8080
  ```

I think in this case the `CONTAINER_PORT` environment variable is wrong. This should point to the oauth-proxy's container port and not Cryostat's container port.

Also, should Cryostat be binding to localhost instead of 0.0.0.0 now?

andrewazores:
I'm working on that ClusterIP setup and related things in #122 . I think the Route and ClusterIP setups in that draft are mostly working but I still need to figure out the other options (Ingress etc.)

andrewazores:
> Also, should Cryostat be binding to localhost instead of 0.0.0.0 now?

That's a good call, I'll put that change here since it's just a oneliner.

ebaron (approved):

ebaron:
Looks good. The remaining issues can be tackled in #122. Good work @aali309!

PR #124, opened 2024-03-08
feat(cryostat3): deploy Cryostat 3.0

andrewazores:
Includes:
- [x] #111 
- [x] #118
- [x] #121 
- [x] #122
- [x] #125
- [x] #127
- [x] #128
- [x] #130 
- [x] #131 
- [x] #132

PR #127, opened 2024-03-18
feat(auth): optionally deploy OpenShift OAuth Proxy

### LINKED ISSUE #117: [Task] Add switch to deploy openshift-oauth-proxy instead of oauth2-proxy
After #114 (and #116), https://github.com/openshift/oauth-proxy/ can be deployed instead of oauth2-proxy. This should also respect the Basic auth configuration from #116 and its `htpasswd` file, if the user provides this configuration. The container should be dropped in in place of oauth2-proxy in all instances and should be configured to pass requests to the same containers in the same manners. The only effective difference to the user should be that the login screen looks different and the credentials are OpenShift account credentials rather than htpasswd Basic credentials.

andrewazores:
Replaces #126
Based on #124
Fixes #117

Not handled here: RBAC. This puts the OpenShift OAuth Proxy in front of the other containers in the Pod, but it will allow any OpenShift account (or Basic credentials if configured additionally) through. Restricting access to OpenShift accounts with particular roles/permissions is still TODO - I wonder if the required set of permissions should just be hardcoded or if this should be something that the deploying user can configure as well.


andrewazores:
Not sure if I'm missing something obvious but with this setup, the OpenShift login redirect flow works and the user can get into the Cryostat Web UI, load all the assets, and make normal HTTP requests. However, the WebSocket connection to the server fails with an `HTTP 500`. As far as I can tell that's coming from the OpenShift OAuth Proxy itself, but it doesn't provide anything in the logs more detailed than simply the access log line indicating that it saw the request and responded with a 500 within 1ms.

andrewazores:
Screenshot_2024-03-18_15-24-08

andrewazores:
I also tried with a rebuild of the oauth-proxy using golang 1.22.1, but it looks the exact same.

```
$ helm install cryostat --set authentication.openshift.enabled=true --set core.route.enabled=true --set openshiftOauthProxy.image.repository=quay.io/andrewazores/openshift-oauth-proxy --set openshiftOauthProxy.image.tag=test-14 ./charts/cryostat/
```

andrewazores:
Other than the WebSocket bug, which I'm pretty well convinced is another openshift-oauth-proxy bug, I think this PR is basically complete.

andrewazores:
If I apply the same interceptor trick to the oauth-proxy rebuild that I did here:

https://github.com/openshift/oauth-proxy/issues/272#issuecomment-1877539666

this causes the proxy to simply pass all requests without actually requiring any authz (no OpenShift redirect etc.), and it also allows the frontend to properly establish its WebSocket connection with the backend via the proxy. That seems to confirm that this is a proxy bug.

andrewazores:
Aha: https://github.com/openshift/oauth-proxy/issues/157

Let me try this.

andrewazores:
Ha. That's incredibly silly.

Screenshot_2024-03-18_16-24-06

charts/cryostat/templates/deployment.yaml
```diff
@@ -300,5 +270,11 @@ spec:
       {{- if .Values.authentication.basicAuth.enabled }}
       - name: {{ .Release.Name }}-htpasswd
         secret:
+          defaultMode: 420
```
ebaron, line 273: This is 0644 isn't it? Should we have something like 0440 instead?

charts/cryostat/values.yaml
```diff
@@ -178,6 +181,25 @@ oauth2Proxy:
     ## @skip oauth2Proxy.securityContext.allowPrivilegeEscalation
     allowPrivilegeEscalation: false
     ## @skip oauth2Proxy.securityContext.capabilities
+    capabilities:
+      drop:
+        - ALL
+          #
```
ebaron, line 187: Is this `#` a typo?

ebaron:
> @ebaron I remember discussing at some point before an idea about unifying the OpenShift SSO support and Route support into one "OpenShift integration" flag, or more generally something like a platform-integration: 'openshift' value. Do you think that would be worthwhile, and if so should I go ahead and do that in this PR? Right now these two features are split into basic.route.enabled and authentication.openshift.enabled.

I think leaving them separate is fine for now. When releasing downstream, we can default them both to true. I imagine the downstream release will typically be how this Helm Chart is used on OpenShift.

> Restricting access to OpenShift accounts with particular roles/permissions is still TODO - I wonder if the required set of permissions should just be hardcoded or if this should be something that the deploying user can configure as well.

It seems like the OAuth proxy expects a JSON string argument with the permissions. We could just define a default and allow the user to override it with their own JSON string. Not the most user-friendly, but the alternative I can think of is supplying a Role or ClusterRole, and at least the latter requires elevated permissions to do.

charts/cryostat/values.yaml
andrewazores: Yup, looks like it, whoops.

andrewazores:
```bash
$ cat htpasswd.conf 
user:$2y$05$Ju.w7ibNuQ/xaAfKH74lJOJ5xqZbONaNWf7HnrjQuZP.4nhLrr0uS

$ oc create secret generic basicauth --from-file=htpasswd.conf
secret/basicauth created

$ helm install cryostat --set authentication.openshift.enabled=true --set core.route.enabled=true --set openshiftOauthProxy.image.repository=quay.io/andrewazores/openshift-oauth-proxy --set openshiftOauthProxy.image.tag=test-14 --set authentication.basicAuth.enabled=true --set authentication.basicAuth.secretName=basicauth --set authentication.basicAuth.filename=htpasswd.conf ./charts/cryostat/

```

With this, both OpenShift SSO and htpasswd Basic login seem to work just fine.

image

ebaron (approved):
Awesome!

Issue #119, opened 2024-01-22
[Request] Allow to edit CRYOSTAT_K8S_NAMESPACES


ElfoLiNk:
### Describe the feature

Would be good to be able to edit CRYOSTAT_K8S_NAMESPACES to watch multiple namespaces.

Probabily also Role and RoleBinding needs fixes to support that

andrewazores:
Thanks @ElfoLiNk , that's a good idea for a feature. It will probably be a while before we're able to prioritize working on that, but I'd be very happy to review such a PR :-)

tthvo:
This is now possible after #128 :D

tthvo:
This should be now configurable with the latest `main` and 3.0+ ^^

PR #128, opened 2024-04-04
feat(discovery): configurations for KubeAPI discovery

tthvo:
Related to #119 
Depends on  https://github.com/cryostatio/cryostat3/pull/325

## Description of changes

Added options to configure discovery mechanism.
- JMX port names & numbers
- Watch namespaces

## How to test

WIP

## WIP

- [ ] Ensure rbac resources are properly configured to support multi-namespaces.


charts/cryostat/values.yaml
```diff
@@ -62,6 +62,17 @@ core:
...
+  discovery:
+    kubernetes:
+      ## @param core.discovery.kubernetes.enabled Enables Kubernetes API discovery mechanism
+      enabled: true
+      ## @param core.discovery.kubernetes.namespaces List of namespaces whose workloads the Cryostat application should be permitted to access and profile.
+      namespaces: []
+      ## @param core.discovery.kubernetes.portNames List of port names that the Cryostat application should look for in order to consider a target as JMX connectable
+      portNames: []
+      ## @param core.discovery.kubernetes.portNumbers List of port numbers that the Cryostat application should look for in order to consider a target as JMX connectable
+      portNumbers: []
```
andrewazores, line 75: Maybe for ease/convenience and consistency with the Operator and old releases, this should default to `namespaces: ["."], portNames: ["jfr-jmx"], portNumbers: [9091]`. If these defaults don't work for the user they can easily override them, but they can also leave these defaults and have something working without too much extra work/reading, and it matches with old documentation we have published.

tthvo: Sounds good! Added now :D

charts/cryostat/templates/role.yaml
andrewazores, line 1: I guess this should be renamed `clusterrole.yaml` and likewise, `clusterrolebinding.yaml` for the other one. Just so the names continue to reflect the contents.

tthvo: Oh sure! Updated now! I kept the `rolebinding.yaml` as it is though since its template for creating rolebinding in watch namespaces.

andrewazores (commented):
LGTM, though may need some adjusting after review on https://github.com/cryostatio/cryostat3/pull/325 finishes.

andrewazores:
@ebaron could you take a quick look over this? I think it looks good and it works well with the accompanying https://github.com/cryostatio/cryostat3/pull/325 (which was just merged).

ebaron (commented):
Sorry for the delay in getting to these Helm PRs. This looks good, just wondering if we should use the same kind of DisableBuiltInPort{Names,Numbers} that we settled on in the Operator's CRD: https://github.com/cryostatio/cryostat-operator/blob/b300689a645e3741ccd5468e09acf20b5b9fb4d2/api/v1beta1/cryostat_types.go#L541-L562 Then we don't need to rely on magic values like `.`.

tthvo:
> Sorry for the delay in getting to these Helm PRs. This looks good, just wondering if we should use the same kind of DisableBuiltInPort{Names,Numbers} that we settled on in the Operator's CRD: https://github.com/cryostatio/cryostat-operator/blob/b300689a645e3741ccd5468e09acf20b5b9fb4d2/api/v1beta1/cryostat_types.go#L541-L562 

I guess if on helm side, if users want to disable built-in port {numbers,names}, they can set, for example, `--set core.discovery.kubernetes.portNumbers="{}" --set core.discovery.kubernetes.portNames="{}"`. Would that be nicer than the bool flag?

> Then we don't need to rely on magic values like ..

I think the `.` is used to represent current namespace, right?

ebaron:
> > Sorry for the delay in getting to these Helm PRs. This looks good, just wondering if we should use the same kind of DisableBuiltInPort{Names,Numbers} that we settled on in the Operator's CRD: https://github.com/cryostatio/cryostat-operator/blob/b300689a645e3741ccd5468e09acf20b5b9fb4d2/api/v1beta1/cryostat_types.go#L541-L562
> 
> I guess if on helm side, if users want to disable built-in port {numbers,names}, they can set, for example, `--set core.discovery.kubernetes.portNumbers="{}" --set core.discovery.kubernetes.portNames="{}"`. Would that be nicer than the bool flag? I am good either way though :D

OpenShift has a UI for customizing Helm values on install, think the Operator one but much more primitive. I think this might be hard to represent that way in those kind of UIs. Boolean values are easily rendered as checkboxes.

> 
> > Then we don't need to rely on magic values like ..
> 
> I think the `.` is used to represent current namespace, right?

It makes sense like `.` is used to represent the current directory on a filesystem, but is this convention used in Kubernetes?

tthvo:
> OpenShift has a UI for customizing Helm values on install, think the Operator one but much more primitive. I think this might be hard to represent that way in those kind of UIs. Boolean values are easily rendered as checkboxes.

Ahh good to know! I didn't realize that. I will update.

> It makes sense like . is used to represent the current directory on a filesystem, but is this convention used in Kubernetes?

Actually, I haven't yet come across any in k8s with this syntax. Though, this is implemented in Cryostat 3.0 tho: https://github.com/cryostatio/cryostat3/blob/1d991d1889e1d63b84f171afe913cf7cb0469aad/src/main/java/io/cryostat/discovery/KubeApiDiscovery.java#L504. @andrewazores any thoughts?

tthvo:
> > OpenShift has a UI for customizing Helm values on install, think the Operator one but much more primitive. I think this might be hard to represent that way in those kind of UIs. Boolean values are easily rendered as checkboxes.
> 
> Ahh good to know! I didn't realize that. I will update.

Sorry another question. Since those fields (i.e. namespaces, portNumbers, portNames) are lists, they are rendered as below. If the users want to disable defaults, would it make sense for them to just remove that item from the list. For example, installing cryostat with securityContext capabilities (using helm chart UI on OpenShift).

image

tthvo:
> Actually, I haven't yet come across any in k8s with this syntax. Though, this is implemented in Cryostat 3.0 tho: https://github.com/cryostatio/cryostat3/blob/1d991d1889e1d63b84f171afe913cf7cb0469aad/src/main/java/io/cryostat/discovery/KubeApiDiscovery.java#L504. @andrewazores any thoughts?

I am seeing #132 using `{{ .Release.Namespace }}`. Would this work better?

andrewazores:
> This does mean cryostat3 can just remove its OWN_NAMESPACE concept?

I'd like to keep the `.` implemented on the Cryostat3 container, even if it isn't used in our own Operator or Helm chart. It's a simple enough feature to carry with minimal maintenance burden, but it seems like it would be quite handy for anyone who wants to roll their own Cryostat deployments.

Otherwise, I'm good with the other changes suggested above.

tthvo:
> I'd like to keep the . implemented on the Cryostat3 container, even if it isn't used in our own Operator or Helm chart. It's a simple enough feature to carry with minimal maintenance burden, but it seems like it would be quite handy for anyone who wants to roll their own Cryostat deployments.

Sounds good to me!

ebaron (commented):
I think this is good. @tthvo what do you think about keeping the Role as a Role, and duplicating it in each namespace? I think so far, a namespace admin is able to install Cryostat using the Helm Chart. They may not have permissions to create a ClusterRole.

tthvo:
> I think this is good. @tthvo what do you think about keeping the Role as a Role, and duplicating it in each namespace? I think so far, a namespace admin is able to install Cryostat using the Helm Chart. They may not have permissions to create a ClusterRole.

Oh yes that makes sense! Totally overlooking this aspect :D

tthvo:
Updated now to copy `Role` over to target namespaces instead of `ClusterRole`. Also, removed checks for magic value `.` in watch namespace list. How about now?

tthvo:
Not related to this PR but one suggestion I think we might benefit from is unit/template testing for helm chart. This can help: https://github.com/helm-unittest/helm-unittest

During this PR, it took me a while to see that the `---` (yaml separator) leaked into the above field (e.g. `namespace: myns---`) due to space trimming but chart renders fine. It would be nice to have testing to ease debugging :D

andrewazores (commented):
LGTM other than the one comment typo above

charts/cryostat/templates/deployment.yaml
```diff
@@ -91,6 +91,18 @@ spec:
...
           - name: GRAFANA_DASHBOARD_URL
             value: http://localhost:3000
+          {{- if .Values.core.discovery.kubernetes.enabled }}
+          - name: CRYOSTAT_DISCOVERY_KUBERNETES_ENABLED
+            value: "true"
+          {{- with .Values.core.discovery.kubernetes }}
+          - name: CRYOSTAT_DISCOVERY_KUBERNETES_NAMESPACES
+            value: {{ include "cryostat.commaSepList" (list .namespaces $.Release.Namespace .installNamespaceDisabled) }}
+          - name: CRYOSTAT_DISCOVERY_KUBERNETES_PORT_NAMES
+            value: {{ include "cryostat.commaSepList" (list .portNames "jmx-jfr" .builtInPortNamesDisabled) }}
```
ebaron, line 101: This should be `jfr-jmx`

charts/cryostat/templates/_helpers.tpl
```diff
@@ -133,3 +133,16 @@ Generate or retrieve a default value for cookieSecret.
 {{- $newSecret | quote -}}
 {{- end }}
 {{- end }}
+
+{{/*
+    Get sanitinized list or defaults (if not disabled) as comma-separated list
```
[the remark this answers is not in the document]
tthvo: Oh opps! Thanks :D

charts/cryostat/templates/deployment.yaml
tthvo: Thankss! Updated :D

PR #131, opened 2024-04-12
fix(authproxy): set openshift oauth proxy prefix to match oauth2_proxy prefix

andrewazores:
Fixes https://github.com/cryostatio/cryostat3/issues/365
Depends on https://github.com/cryostatio/cryostat3/pull/366


andrewazores:
Maybe a hackish workaround for the bug above would be to just force a refresh on the web-client side if it looks like there was a request failure?

aali309 (approved):

PR #138, opened 2024-05-13
fix(upgrade): fix errors during chart version upgrade

tthvo:
Related to #124 

## What's changed

- Wrapped new configurations in `()` for nil-checks. See [here](https://stackoverflow.com/questions/59795596/how-to-make-nested-variables-optional-in-helm/68807258#68807258).
  - However, this is just for safe guarding, not solving the upgrade failures. See below.
- Create a wrapper that forces helm to use `--reset-then-reuse-values`. This is available only with helm 3.14+.
  - The user cannot use `helm upgrade --reuse-values` as it would cause values such as `(.Values.storage).image.repository` to be empty. This is an invalid image repository.
  - Helm 3.14+ has a new feature with `--reset-then-reuse-values`: https://github.com/helm/helm/pull/9653 that fits our upgrade path more.
  - However, `chart-testing` has not been updated for accommodate that so this is a temporary fix.

- Removed unused test value files (minimal mode is removed).
- Upgraded related github actions.

## What's motivation

The test is failing because chart-testing `ct` is also testing upgrade path with `--reuse-values`. This causes `nil` pointer error as old values do not have newly defined fields in 3.0.

## Sample run

[Sample PR](https://github.com/tthvo/cryostat-helm/pull/7): https://github.com/tthvo/cryostat-helm/actions/runs/9061712431/job/24894069674?pr=7

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -27,21 +27,22 @@ on:
 
 env:
   TARGET_BRANCH: ${{ github.event.pull_request.base.ref || github.ref_name }}
+  TEST_NAMESPACE: helm-test
 
 jobs:
   lint-chart:
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v2
+      - uses: actions/checkout@v4
         with:
           fetch-depth: 0
       - name: Set up Helm
-        uses: azure/setup-helm@v3
+        uses: azure/setup-helm@v4
         with:
-          version: v3.11.2
-      - uses: actions/setup-python@v4
+          version: v3.14.4
+      - uses: actions/setup-python@v5
         with:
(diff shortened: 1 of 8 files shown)


andrewazores (commented):
Seems like this makes sense.

andrewazores:
@ebaron WDYT?

ebaron (approved):
Thanks for doing this work @tthvo. It's a shame the tester doesn't support this case, since we're not planning on upgrades from 2.x to 3.0 working out of the box in all scenarios.

tthvo:
I am hoping chart-testing would add a patch to support each upgrade strategy after https://github.com/helm/chart-testing/issues/525. If not, I can try in the next couple days as many might run into the same issues as us :D

PR #152, opened 2024-06-12
fix(secret): database secret should be immutable

tthvo:
Related to https://github.com/cryostatio/cryostat-operator/pull/831

## Descriptions

- Refactored to manage only a single secret  for both database connection and encryption key.
- Marked generated database secret as `immutable` -> bumped supported min k8s version to `1.21.0`.

## Motivations

- Following https://github.com/cryostatio/cryostat-operator/pull/831, Database secret should be immutable to avoid accidental changes to secret data.
- This should bring the resources managed by the chart closer those on the operator side.
- This also allows user to customize connection key to database if needed. Previously, it was only possible with encryption key.


andrewazores (approved):
LGTM. @ebaron?

ebaron (approved):
Looks good, thanks @tthvo!

PR #154, opened 2024-06-12
chore(chart): update chart sources

tthvo:
Related to #110 

## Descriptions

Updated list of source code for helm chart: https://helm.sh/docs/topics/charts/#the-chartyaml-file

### DIFF
--- a/charts/cryostat/Chart.yaml
+++ b/charts/cryostat/Chart.yaml
@@ -26,11 +26,13 @@ keywords:
 - diagnostic
 
 sources:
-- https://github.com/cryostatio/cryostat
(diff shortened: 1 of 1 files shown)


andrewazores (commented):
Should probably add https://github.com/cryostatio/openshift-oauth-proxy/ as well, and maybe https://github.com/oauth2-proxy/oauth2-proxy

tthvo:
> Should probably add https://github.com/cryostatio/openshift-oauth-proxy/ as well, and maybe https://github.com/oauth2-proxy/oauth2-proxy

Yup sounds good! Added :D

andrewazores (approved):

Issue #142, opened 2024-05-29
[Request] Unit/Template Tests for helm chart


tthvo:
## Describe the feature

> Not related to this PR but one suggestion I think we might benefit from is unit/template testing for helm chart. This can help: https://github.com/helm-unittest/helm-unittest

> During this PR, it took me a while to see that the `---` (yaml separator) leaked into a template field (e.g. `namespace: myns---`) due to space trimming but chart renders fine. It would be nice to have testing to ease debugging :D

_Originally posted by @tthvo in https://github.com/cryostatio/cryostat-helm/issues/128#issuecomment-2136249999_

## Anything other information?

Down the line, we can establish some more e2e tests like scorecards, on top of the connection test: https://helm.sh/docs/topics/chart_tests/

tthvo:
This requires a bit of research to design a framework and conventions. @ebaron @andrewazores what are your thoughts?

andrewazores:
It's worth investing the time into. After 3.0 is shipped we should continue to spend some time on improving the overall testing situation because 3.0 is quite far behind what 2.4 had, which already didn't have as much automated testing as I would like.

Adding more testing capabilities at higher levels of abstraction, like the Helm chart and Operator scorecard tests, seems like a powerful tool. Something we have always lacked in our automated testing has been the capability to test with different configuration parameters (ie environment variables), so that was only ever done in the old codebase by whitebox unit testing and using mocks to test very specific behaviours under different environment variable settings. Testing at these higher levels lets us do integration/end-to-end testing under those conditions which is very preferable.

tthvo:
Thanks, sounds good! @mwangggg @aali309 anyone interested in this work?

aali309:
> Thanks, sounds good! @mwangggg @aali309 anyone interested in this work?

Yes, I can try to look at it and see how far I can go once done with my pending PRs.

tthvo:
I guess this can now be closed?

Issue #115, opened 2024-01-18
[Task] Place auth-proxy in front of split storage and database deployments


andrewazores:
See https://github.com/cryostatio/cryostat-operator/wiki/Security-and-Authz#scenario , #114

andrewazores:
Both the Storage and DB containers are already set up with simple authentication using generated Secrets. Adding an oauth2 proxy in Basic mode in front to enforce authentication doesn't seem like it adds anything new. Using the OpenShift oauth proxy in front to use the Cryostat serviceaccount instead of generated credentials might be nice, but probably only works for the HTTP-based Storage and adds a hard dependency on OpenShift.

Using kube-rbac-proxy instead for authentication and k8s serviceaccount authorization would be nice, but this also seems like it likely only works for HTTP-based storage and not the JDBC database. Also, kube-rbac-proxy requires TLS configuration, so this would mean either we require the user to supply TLS configuration via Secrets, or else we have to add a dependency on something like cert-manager to automate this.

Related: #168

andrewazores:
There are actually TLS-related Helm functions available:

https://helm.sh/docs/chart_template_guide/function_list/#cryptographic-and-security-functions

These look like a reasonable alternative to cert-manager for the Helm chart's purposes.

andrewazores:
Closing - the remaining viable ideas here are a duplicate of #168.

PR #159, opened 2024-06-26
feat(Test): template tests

aali309:
see  [https://github.com/cryostatio/cryostat-helm/issues/142](https://github.com/cryostatio/cryostat-helm/issues/142)


aali309:
Working on `TESTING.md`  to serve as a guide for contributors.

tthvo:
@aali309 We should also create a workflow to run these tests on PR and push. What do you think?

aali309:
> @aali309 We should also create a workflow to run these tests on PR and push. What do you think?

Will do that as well

tthvo (commented):
Looks like a good start! I think there are still cases and validations that need covering.

As for the `NOTES.txt`, you can do the following, for example:

```yaml
suite: Test NOTES.txt
templates:
  - templates/NOTES.txt

tests:
  - it: should be render correct notes for enabled ingress
    set:
      core.ingress.enabled: true
      core:
        ingress:
          className: nginx
          hosts:
            - host: example.com
              paths:
                - path: /
                  pathType: Prefix
          tls:
            - hosts:
                - example.com
              secretName: example-tls
    asserts:
      - equalRaw:
          value: |


            1. Visit the Cryostat application at:
              ```
              https://example.com/
              ```
```

charts/cryostat/tests/clusterrolebinding_test.yaml
```diff
@@ -0,0 +1,65 @@
+suite: test ClusterRoleBinding
```
tthvo, line 1: I would also recommend having a convention to naming test suite :D

tthvo: Same for test cases ^^ But seems like they pretty much consistent already.

andrewazores:
This looks really good, seems like a great and lightweight way to define some nice regression tests and test Helm chart logic quickly. Are there any more test cases planned or in progress at this point or is it ready for review?

aali309:
> This looks really good, seems like a great and lightweight way to define some nice regression tests and test Helm chart logic quickly. Are there any more test cases planned or in progress at this point or is it ready for review?

Its ready for review. While reviewing if more test cases are needed I can add them.

charts/cryostat/tests/deployment_test.yaml
```diff
@@ -0,0 +1,164 @@
...
+    asserts:
+      - equal:
+          path: spec.template.spec.containers[?(@.name=='cryostat')].imagePullPolicy
+          value: Always
+      - equal:
+          path: spec.template.spec.containers[?(@.name=='cryostat-authproxy')].imagePullPolicy
+          value: Always
+      - equal:
+          path: spec.template.spec.containers[?(@.name=='cryostat-storage')].imagePullPolicy
+          value: Always
```
andrewazores, line 151: The pull policies get set to `IfNotPresent` for release versions, but `Always` in between for development snapshots. Could that behaviour be accounted for in this test?

charts/cryostat/tests/deployment_test.yaml
```diff
@@ -0,0 +1,164 @@
...
+      - equal:
+          path: metadata.labels
+          value:
+            helm.sh/chart: cryostat-0.5.0-dev
+            app.kubernetes.io/name: cryostat
+            app.kubernetes.io/instance: RELEASE-NAME
+            app.kubernetes.io/version: "4.0.0-dev"
+            app.kubernetes.io/managed-by: Helm
+
+  - it: should validate all container settings including environment variables and security contexts
```
tthvo, line 29:
```suggestion
  - it: should validate all container environment variables and health probes
```

Does not seem to validate security context? There is another test case below for that.

charts/cryostat/tests/deployment_test.yaml
```diff
@@ -0,0 +1,164 @@
...
+
+  - it: should ensure the datasource container has correct environment variables and resources
+    asserts:
+      - equal:
+          path: spec.template.spec.containers[?(@.name=='cryostat-jfr-datasource')].env[?(@.name=='QUARKUS_HTTP_PORT')].value
+          value: "8800"
+      - exists:
+          path: spec.template.spec.containers[?(@.name=='cryostat-jfr-datasource')].resources
+
+  - it: should validate security contexts for each container
```
tthvo, line 130: We say `each container` but only core container is verified? We should include others.

charts/cryostat/tests/deployment_test.yaml
```diff
@@ -0,0 +1,164 @@
...
+        effect: "NoSchedule"
+    asserts:
+      - exists:
+          path: spec.template.spec.nodeSelector
+      - exists:
+          path: spec.template.spec.affinity
+      - exists:
+          path: spec.template.spec.tolerations
+
+  - it: should ensure the datasource container has correct environment variables and resources
```
tthvo, line 122: If we are setting up the tests based on setting types (i.e. security context, env vars, pull policies), then each test should validate all containers. So this can be merged with above test case: `should validate all container settings including environment variables...`

charts/cryostat/tests/role_test.yaml
```diff
@@ -0,0 +1,141 @@
...
+
+  - it: should not create any Roles when Kubernetes discovery is disabled
+    set:
+      rbac.create: true
+      core.discovery.kubernetes.enabled: false
+    asserts:
+      - hasDocuments:
+          count: 0
+
+  - it: should not specify a namespace if installNamespaceDisabled is true
```
tthvo, line 122:
```suggestion
  - it: should create a role in install namespace by default if installNamespaceDisabled is true
```

charts/cryostat/tests/role_test.yaml
```diff
@@ -0,0 +1,141 @@
...
+    set:
+      rbac.create: true
+      core.discovery.kubernetes.enabled: true
+      core.discovery.kubernetes.installNamespaceDisabled: true
+    asserts:
+      - hasDocuments:
+          count: 0
+
+  - it: should specify the default namespace if installNamespaceDisabled is false
```
tthvo, line 131:
```suggestion
  - it: should create a role in install namespace by default if installNamespaceDisabled is false
```

charts/cryostat/tests/rolebinding_test.yaml
```diff
@@ -0,0 +1,94 @@
...
+      - equal:
+          path: subjects[0].namespace
+          value: "NAMESPACE"
+        documentIndex: 0
+      - equal:
+          path: subjects[0].namespace
+          value: "NAMESPACE"
+        documentIndex: 1
+
+  - it: should not create any RoleBinding if RBAC is disabled
```
tthvo, line 89: Missing cases for discovery options as in  `role_test.yaml` (we should basically mirror those cases over here)?

charts/cryostat/tests/notes_test.yaml
```diff
@@ -0,0 +1,90 @@
...
+
+            1. Visit the Cryostat application at:
+              ```
+              https://example.com/
+              ```
+
+  - it: should render notes for enabled route without TLS
+    set:
+      core.route.enabled: true
+      core.route.tls.enabled: false
```
tthvo, line 34: What about tls enabled?

charts/cryostat/tests/service_test.yaml
```diff
@@ -0,0 +1,74 @@
...
+          value: 443
+      - equal:
+          path: spec.ports[1].targetPort
+          value: 8443
+      - equal:
+          path: spec.ports[1].protocol
+          value: TCP
+      - equal:
+          path: spec.ports[1].name
+          value: cryostat-https
```
tthvo, line 46: Hey @ebaron @andrewazores, I notice here we set the https port even though we don't have TLS setup yet. Is it okay to do so?

andrewazores: I guess it's "okay" depending on how the Route or Ingress are configured. If TLS is terminated at the edge rather than passed through then this probably gets effectively ignored anyway. The Route is always created pointing at service port 4180, which is the HTTP port, so this only makes sense when either using no TLS or using edge termination on the Route.

This means right now there might be some ways to get an invalid setup which has halfway-configured TLS and in the end traffic won't actually flow.

tthvo: > This means right now there might be some ways to get an invalid setup which has halfway-configured TLS and in the end traffic won't actually flow.

```
NAME         TYPE        CLUSTER-IP    EXTERNAL-IP   PORT(S)            AGE     SELECTOR
cryostat     ClusterIP   10.99.43.93   <none>        8181/TCP,443/TCP   3m20s   app.kubernetes.io/instance=cryostat,app.kubernetes.io/name=cryostat
```
I guess it's just odd that the k8s Service shows a 443 port available, but when you port-foward it, the connection is refused. Maybe, we can remove it for now?

andrewazores: Forwarding that port would be one case of such an invalid setup :-)

We can remove this port definition for now, since it depends on getting the auth proxies configured to use TLS first before it can actually do anything.

tthvo: Sounds good. #167 :D

ebaron (commented):
Fantastic work @aali309!

aali309:
I will be adding more test cases on this.

charts/cryostat/tests/service_test.yaml
```diff
@@ -0,0 +1,62 @@
...
+          path: spec.ports[0].name
+          value: cryostat-http
+      - equal:
+          path: metadata.labels
+          value:
+            app.kubernetes.io/instance: RELEASE-NAME
+            app.kubernetes.io/managed-by: Helm
+            app.kubernetes.io/name: cryostat
+            app.kubernetes.io/version: "4.0.0-dev"
+            helm.sh/chart: cryostat-0.5.0-dev
```
tthvo, line 42: Hey @andrewazores, I am seeing Chart version is still `0.5.0-dev` on `main`. Should this be something `1.x`?

andrewazores: Nice catch, yes, probably `1.1`? @ebaron any other idea?

ebaron: 1.1 or 2.0 are fine with me. No strong feeling either way.

tthvo: I guess if we are setting `4.0.0-dev` on the operator side, we can use `2.0` here since there can be upcoming breaking changes (e.g. #115)?

aali309:
ping @andrewazores @ebaron

.github/workflows/unittest.yml
```diff
@@ -0,0 +1,41 @@
...
+        run: exit 1
+      - uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+      - name: Set up Helm
+        uses: azure/setup-helm@v4
+        with:
+          version: v3.14.4
+      - name: Install unit test plugin
+        run: helm plugin install https://github.com/helm-unittest/helm-unittest.git
```
andrewazores, line 39: Maybe this should have a `--version` so we can pin it to a version we know works?

charts/cryostat/TESTING.md
```diff
@@ -0,0 +1,148 @@
...
+          path: <path to value to test>
+          value: <expected value>
+```
+## Common Assertions
+- `equal`: Checks if the actual value at path equals the expected value.
+- `matchRegex`: Validates if the actual string matches the given regex pattern.
+- `exists`: Checks if the specified path exists in the document.
+- `notExists`: Ensures the specified path does not exist in the document.
+
+Visit [ for more Assertion Types](https://github.com/helm-unittest/helm-unittest/blob/main/DOCUMENT.md#assertion-types).
```
andrewazores, line 111:
```suggestion
See [this document](https://github.com/helm-unittest/helm-unittest/blob/main/DOCUMENT.md#assertion-types) for more assertion types.
```

andrewazores:
Looks great! Just the two tiny notes above.

charts/cryostat/tests/pvc_test.yaml
```diff
@@ -0,0 +1,56 @@
...
+          path: metadata.name
+          value: "RELEASE-NAME-cryostat"
+      - equal:
+          path: metadata.labels
+          value:
+            app.kubernetes.io/instance: RELEASE-NAME
+            app.kubernetes.io/managed-by: Helm
+            app.kubernetes.io/name: cryostat
+            app.kubernetes.io/version: "4.0.0-dev"
+            helm.sh/chart: cryostat-0.5.0-dev
```
andrewazores, line 35: I guess there are more of these that need updating now too.

charts/cryostat/tests/service_test.yaml
```diff
@@ -0,0 +1,62 @@
...
+          value: 8080
+      - equal:
+          path: spec.ports[0].targetPort
+          value: 4180
+      - equal:
+          path: spec.ports[0].protocol
+          value: TCP
+      - equal:
+          path: spec.ports[0].name
+          value: cryostat-http
```
tthvo, line 34: We also check if `cryostat-https` port is available when openshift auth is enabled.

tthvo: See #167

tthvo (commented):
Looks good to me! This is awesome!

andrewazores (approved):

PR #160, opened 2024-06-27
chore(templates): ensure named templates follow helm best practices

tthvo:
Related to #110 
Related to #159  (some are required for helm unittest to function, i.e. point [1] and [2])

## What's changed

Changes that follow best practices:

- Renamed files ending with `.tpl` to have underscore prefix `_`. [1]
- Renamed named templates with chart name (i.e. named templates are global so its best practice to prefix with chart name to avoid collision in case sub-charts are available).
- Added comments to named templates if missing.

Changes that might impact chart installation:
- Removed duplicate `securityContext` field for storage container. [2]
- Fixed incorrect indentation for securityContext field for proxy containers.
- Fixed nil pointer exception when generating `NOTEs.txt` with ingress enabled.

## References

- https://helm.sh/docs/chart_best_practices/templates/
- https://helm.sh/docs/chart_template_guide/named_templates/


andrewazores (approved):

ebaron (approved):
Thanks @tthvo! Seems like there are some pretty important bugfixes here. You think we should do a last minute backport to 3.0?

andrewazores:
Seems worthwhile.

PR #164, opened 2024-07-03
feat(ci): added unittest workflow

aali309:
suggested as part of: #159

### DIFF
--- a/.github/workflows/unittest.yml
+++ b/.github/workflows/unittest.yml
@@ -1,4 +1,4 @@
-name: Helm Unittests
+name: Helm unit tests
 
 on:
   push:
@@ -25,19 +25,18 @@ jobs:
   helm-unittests:
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v4
-        with:
-          fetch-depth: 0
       - name: fail if safe-to-test label NOT applied
         if: ${{ github.event_name == 'pull_request' && !contains(github.event.pull_request.labels.*.name, 'safe-to-test') }}
         run: exit 1
       - uses: actions/checkout@v4
-      
+        with:
+          fetch-depth: 0
+
       - name: Set up Helm
         uses: azure/setup-helm@v4
         with:
           version: v3.15.2
-      - name: Install Unittest Plugin
+      - name: Install unit test plugin
         run: helm plugin install https://github.com/helm-unittest/helm-unittest.git
-      - name: Run Helm Unittests
+      - name: Run Helm unit tests
         run: helm unittest --debug ./charts/cryostat
--- a/.github/workflows/unittest.yml
+++ b/.github/workflows/unittest.yml
@@ -1,0 +1,43 @@
+name: Helm Unittests
+
+on:
+  push:
+    branches:
+      - main
+      - v[0-9]+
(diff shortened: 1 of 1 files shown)


aali309:
link: [https://github.com/aali309/cryostat-helm/actions/runs/9781866183/job/27006928374](https://github.com/aali309/cryostat-helm/actions/runs/9781866183/job/27006928374)

Sample local run:
image

.github/workflows/unittest.yml
```diff
@@ -0,0 +1,43 @@
+name: Helm Unittests
```
mwangggg, line 1: Probably want to separate Unit Tests into two words for clarity (here and elsewhere)

.github/workflows/unittest.yml
```diff
@@ -0,0 +1,43 @@
...
+      - cryostat-v[0-9]+.[0-9]+
+
+jobs:
+  helm-unittests:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+      - name: fail if safe-to-test label NOT applied
```
mwangggg, line 31: we'll probably want this step to be first, before the checkout action

.github/workflows/unittest.yml
```diff
@@ -0,0 +1,42 @@
...
+        if: ${{ github.event_name == 'pull_request' && !contains(github.event.pull_request.labels.*.name, 'safe-to-test') }}
+        run: exit 1
+      - uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+
+      - name: Set up Helm
+        uses: azure/setup-helm@v4
+        with:
+          version: v3.15.2
```
tthvo, line 38: We are using `v3.14.4` in other CI. Let's keep it at `3.14` for now and bump later.

tthvo (commented):
Looks good! I would suggest having these changes included in #159 instead so that we can see the tests run in that PR too.

.github/workflows/unittest.yml
```diff
@@ -0,0 +1,42 @@
...
+  helm-unittests:
+    runs-on: ubuntu-latest
+    steps:
+      - name: fail if safe-to-test label NOT applied
+        if: ${{ github.event_name == 'pull_request' && !contains(github.event.pull_request.labels.*.name, 'safe-to-test') }}
+        run: exit 1
+      - uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+
```
mwangggg, line 34: this blank line can be removed

.github/workflows/unittest.yml
aali309: or may be change `v3.14.4` to match the latest?

tthvo: Sure that would work! Though, it's better to bump the version in a separate PR.

tthvo: > If something goes wrong with fetching the latest version the action will use the hardcoded default stable version (currently v3.13.3). If you rely on a certain version higher than the default, you should explicitly use that version instead of latest.


We need to be careful there since helm `3.13` might be used, while the min helm version required is `3.14`.

PR #166, opened 2024-07-03
chore(release): set versions for 3.0.1 development

andrewazores:
See #165


ebaron (commented):
Some of these images use a `3.0` tag. Are we expecting this to refer to the latest development snapshot, or the latest 3.0.x release?

andrewazores:
Ah right, I think those `3.0`s should all be `3.0-dev`s:

https://github.com/cryostatio/cryostat-db/blob/dde25a293cb2e2b2b39e2c7a0f11727387ef0559/.github/workflows/ci.yml#L39

andrewazores:
Maybe not: https://quay.io/repository/cryostat/cryostat-storage?tab=tags

It doesn't look like that tag is actually getting published the way I expected. So this should actually be `cryostat-v3.0`?

ebaron (commented):
Just needs another run of readme-generator-for-helm

ebaron:
> Maybe not: https://quay.io/repository/cryostat/cryostat-storage?tab=tags
> 
> It doesn't look like that tag is actually getting published the way I expected. So this should actually be `cryostat-v3.0`?

That would make sense

ebaron (approved):

PR #167, opened 2024-07-09
chore(service): https port should not be exposed on k8s environment

tthvo:
Related to #159 
Related to #127

## Descriptions

Un-exposed invalid port definition in Service (i.e. https port) that is not yet implemented on non-OpenShift environment.

```
NAME         TYPE        CLUSTER-IP      EXTERNAL-IP   PORT(S)    AGE    SELECTOR
cryostat     ClusterIP   10.110.17.206   <none>        8181/TCP   55s    app.kubernetes.io/instance=cryostat,app.kubernetes.io/name=cryostat
```

## Motivations

https://github.com/cryostatio/cryostat-helm/pull/159#discussion_r1671083612

### DIFF
--- a/charts/cryostat/templates/service.yaml
+++ b/charts/cryostat/templates/service.yaml
@@ -17,9 +17,11 @@ spec:
       targetPort: 4180
       protocol: TCP
       name: cryostat-http
+    {{- if (.Values.authentication.openshift).enabled }}
     - port: 443
       targetPort: 8443
       protocol: TCP
       name: cryostat-https
+    {{- end }}
   selector:
     {{- include "cryostat.selectorLabels" $ | nindent 4 }}


andrewazores:
Hmm, actually:

https://github.com/cryostatio/cryostat-helm/blob/8115c826b8892a08a1232b1aab5b3c54a4157d6d/charts/cryostat/templates/_openshiftOauthProxy.tpl#L21

8443 is used for HTTPS when the OpenShift OAuth Proxy is deployed, since this setup is clever enough to automatically pick up and use OpenShift TLS certs.

https://github.com/cryostatio/cryostat-helm/blob/8115c826b8892a08a1232b1aab5b3c54a4157d6d/charts/cryostat/templates/service.yaml#L11

https://github.com/cryostatio/cryostat-helm/blob/8115c826b8892a08a1232b1aab5b3c54a4157d6d/charts/cryostat/templates/deployment.yaml#L277

But for non-OpenShift, or when deploying the OAuth2 Proxy on OpenShift, then TLS is not yet supported and this port doesn't make sense.

So maybe this port should be conditionally added only if `(.Values.authentication.openshift).enabled`?

tthvo:
Ahh that makes sense! Let me update this PR. I primarily use Cryostat with k8s lately :D

andrewazores (approved):

PR #173, opened 2024-07-16
ci(oci): push helm chart package to github container registry

mwangggg:
fixes: #161

PR #176, opened 2024-07-17
ci(oci): attempt to push using only registry/namespace

andrewazores:
See #175 #174 #173

### DIFF
--- a/.github/workflows/push-oci.yml
+++ b/.github/workflows/push-oci.yml
@@ -33,7 +33,6 @@ jobs:
         uses: helm/chart-releaser-action@v1.6.0
         env:
           CR_TOKEN: "${{ secrets.GITHUB_TOKEN }}"
-          CR_SKIP_EXISTING: "true"
         with:
           charts_dir: charts
           config: .github/helm-release-config.yml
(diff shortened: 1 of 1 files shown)


tthvo:
We might clash with https://github.com/cryostatio/cryostat-legacy/pkgs/container/cryostat?

andrewazores:
Yea, that's what happens: https://github.com/cryostatio/cryostat-legacy/pkgs/container/cryostat

Need some way to get the chart releaser to package the chart with `cryostat-helm` as its name, not just `cryostat`.

andrewazores:
Or actually, I guess it's `helm push` which is giving us problems, not the chart releaser.

tthvo:
The chart releaser is also having issue I think since the release still got published. Can we just use `helm package`?

andrewazores:
Maybe we can just hack it by patching the `charts/cryostat/Chart.yaml` to set `name: cryostat-helm` within the CI run...

tthvo:
https://trstringer.com/helm-charts-github-container-registry/

tthvo:
> Maybe we can just hack it by patching the charts/cryostat/Chart.yaml to set name: cryostat-helm within the CI run...

That sounds reasonable to me :D

tthvo:
Also, I figured out the `skip_upload`. Seems like it is only available on `main`, not latest `1.6.0` :D

PR #177, opened 2024-07-17
ci(oci): helm repackaging

andrewazores:
See #176 #173

PR #178, opened 2024-07-17
chore(notes): simplify post-install steps for ClusterIP service case

tthvo:
Related to #177, #160 

## Descriptions

- Simplify post-install steps for ClusterIP service case.
- Use generic word `application` to refer to Cryostat in release notes. I suppose installing Cryostat chart already means installing Cryostat (See https://github.com/cryostatio/cryostat-helm/pull/177#issuecomment-2234258284)

## Motivations

When installing the chart without further customizations, the user just needs to port-forward the local port to the application pod. Previously, there were some extra steps that only defined some environment variables (i.e. not actually configuring cryostat).

## How to test

```
helm install cryostat-sample charts/cryostat
```

Expected notes:

````
NOTES:

1. Configure the application to serve external traffic:
  ```
  No actions required with this configuration.
  ```

2. Forward local port to the application's pod:
  ```
  kubectl -n default wait --for=condition=available --timeout=60s deploy/cryostat-sample

  export POD_NAME=$(kubectl get pods -n default -l "app.kubernetes.io/name=cryostat,app.kubernetes.io/instance=cryostat-sample" --sort-by=.metadata.creationTimestamp -o jsonpath="{.items[-1:].metadata.name}")
  export CONTAINER_PORT=$(kubectl get pod -n default $POD_NAME -o jsonpath="{.spec.containers[0].ports[0].containerPort}")
  kubectl -n default port-forward $POD_NAME "8080:$CONTAINER_PORT"
  ```

3. Visit the application at:
  ```
  http://localhost:8080
  ```
````

### DIFF
--- a/charts/cryostat/templates/NOTES.txt
+++ b/charts/cryostat/templates/NOTES.txt
@@ -52,7 +52,7 @@
 {{- else if .Values.core.ingress.enabled }}
 {{- range $host := .Values.core.ingress.hosts -}}
   {{- range .paths }}
-  echo {{ ternary "http" "https" (empty $.Values.core.ingress.tls) }}://{{ $host.host }}{{ .path }}
+  {{ ternary "http" "https" (empty $.Values.core.ingress.tls) }}://{{ $host.host }}{{ .path }}
   {{- end }}
 {{- end }}
 {{- else if contains "NodePort" .Values.core.service.type }}
@@ -60,6 +60,6 @@
 {{- else if contains "LoadBalancer" .Values.core.service.type }}
   echo http://$SERVICE_IP:{{ .Values.core.service.httpPort }}
 {{- else if contains "ClusterIP" .Values.core.service.type }}
-  echo http://localhost:8080
+  http://localhost:8080
 {{- end }}
   ```
--- a/charts/cryostat/templates/NOTES.txt
+++ b/charts/cryostat/templates/NOTES.txt
@@ -52,7 +52,7 @@
 {{- else if .Values.core.ingress.enabled }}
 {{- range $host := .Values.core.ingress.hosts -}}
   {{- range .paths }}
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by tthvo** [comment]
I did take the opportunity to clean up README to make it a little nicer :D https://github.com/tthvo/cryostat-helm/blob/release-notes/charts/cryostat/README.md

File: charts/cryostat/templates/NOTES.txt [hunk 1]
```diff
@@ -32,32 +32,34 @@
...
-{{ $listNum }}. {{ "Visit the " }}{{ .Chart.Name | camelcase }}{{ " application at: " }}
+{{ $listNum }}. Visit the application at:
   ```
 {{- if .Values.core.route.enabled }}
   echo {{ ternary "https" "http" .Values.core.route.tls.enabled }}://$(oc get route -n {{ .Release.Namespace }} {{ include "cryostat.fullname" . }} -o jsonpath="{.status.ingress[0].host}")
 {{- else if .Values.core.ingress.enabled }}
 {{- range $host := .Values.core.ingress.hosts -}}
   {{- range .paths }}
-  {{ ternary "http" "https" (empty $.Values.core.ingress.tls) }}://{{ $host.host }}{{ .path }}
+  echo {{ ternary "http" "https" (empty $.Values.core.ingress.tls) }}://{{ $host.host }}{{ .path }}
```
andrewazores, line 55: The `echo` doesn't seem necessary here since all the substitutions are being done by helm - on line 51 the `echo` is needed because there is a shell command `$(oc get route)` emitted as part of the instructions. I guess it seems nice to make these all look similar with an `echo` in front, but I'd rather only put that there if it's actually needed for something.

tthvo: Ah yes, make sense! I guess that applies for other cases where environment variables are present. Thanks!

andrewazores (approved):

PR #179, opened 2024-07-18
ci(lint): reorganize workflows and add jobs to check readme diff

tthvo:
Related to #73 

### Chore
- Split Chart Lint and Test into separate workflow.
- Rename workflows to be consistent.
- Cleaned up README Installation section to be a bit nicer.

### Feature

- Added a job to check if README requires updating.

## Motivations

Currently, developers have to remember regenerating README + using the correct readme-generator tool and version. This new workflow can serve as a reminder and suggest hints on which command to run to update README.

## Others

The latest version of [readme-generator](https://github.com/bitnami/readme-generator-for-helm) (`2.6.1`) has some breaking changes, I believe. So, I am keeping it as `2.5.x`.

## Samples

- Failed (needs regenerating): https://github.com/tthvo/cryostat-helm/actions/runs/9987975619/job/27603459670?pr=8
- Succeeded: https://github.com/cryostatio/cryostat-helm/actions/runs/9987706169/job/27602625206?pr=179


.github/workflows/chart-lint.yml
```diff
@@ -0,0 +1,70 @@
+name: Lint Helm Chart
+
+concurrency:
+  group: ci-${{ github.run_id }}
+  cancel-in-progress: true
```
mwangggg, line 5: we should probably add concurrency to the `push-ci.yml` file too?

.github/workflows/chart-lint.yml
```diff
@@ -0,0 +1,70 @@
...
+      - v[0-9]+
+      - v[0-9]+.[0-9]+
+      - cryostat-v[0-9]+.[0-9]+
+
+jobs:
+  lint-chart:
+    runs-on: ubuntu-latest
+    env:
+      TARGET_BRANCH: ${{ github.event.pull_request.base.ref || github.ref_name }}
+      TEST_NAMESPACE: helm-test
```
mwangggg, line 33: I don't think this variable is needed in this job

tthvo: nice catch! Thanks!

.github/workflows/chart-lint.yml
tthvo: make sense! Added now!

andrewazores (approved):

PR #180, opened 2024-08-07
fix(authproxy): cookie secret creation and reference using env valueFrom

elias-gb:
### Fix

- Added relevant cookie secret.
- Added test for cookie secret.

## Motivations 
There was no cookie secret to be looked up in subsequent helm upgrade.

### DIFF
--- a/charts/cryostat/templates/cookie_secret.yaml
+++ b/charts/cryostat/templates/cookie_secret.yaml
@@ -1,0 +1,7 @@
+apiVersion: v1
+kind: Secret
+metadata:
+  name: {{ .Release.Name }}-cookie-secret
+type: Opaque
+data:
+  COOKIE_SECRET: {{ include "cryostat.cookieSecret" . }}
--- a/charts/cryostat/templates/cookie_secret.yaml
+++ b/charts/cryostat/templates/cookie_secret.yaml
@@ -1,7 +1,13 @@
+{{- $secret := (lookup "v1" "Secret" .Release.Namespace (printf "%s-cookie-secret" .Release.Name)) -}}
 apiVersion: v1
 kind: Secret
 metadata:
   name: {{ .Release.Name }}-cookie-secret
 type: Opaque
+{{ if $secret -}}
+data:
+  COOKIE_SECRET: {{ $secret.data.COOKIE_SECRET }}
+{{ else -}}
 stringData:
   COOKIE_SECRET: {{ randAlphaNum 32 }}
+{{ end }}
--- a/charts/cryostat/templates/cookie_secret.yaml
+++ b/charts/cryostat/templates/cookie_secret.yaml
@@ -3,5 +3,5 @@ kind: Secret
 metadata:
   name: {{ .Release.Name }}-cookie-secret
 type: Opaque
-data:
-  COOKIE_SECRET: {{ include "cryostat.cookieSecret" . }}
+stringData:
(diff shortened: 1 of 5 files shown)


.gitignore
```diff
@@ -0,0 +1 @@
+.idea
```
andrewazores, line 1: Please configure yourself a global gitignore on your workstation, and put these IDE-specific files and directories there instead.

andrewazores (commented):
Hi,

Thanks for the contribution! Could you detail how to reproduce the bug you ran into? It sounds like you did an install and then an upgrade, and that's when you hit a problem with the Secret?

elias-gb:
> Hi,
> 
> Thanks for the contribution! Could you detail how to reproduce the bug you ran into? It sounds like you did an install and then an upgrade, and that's when you hit a problem with the Secret?

There is no secrets created for cookies during both the initial install and upgrades. We noticed the issue as you described when we tried to install and upgrade. Each time it tried to recreate the cookie secrets on authproxy.

andrewazores:
@elias-gb could you test the latest commit I pushed and ensure it works on your setup?

The Secret you've defined should get properly created on installation still, but it will get recreated with a new secret value on upgrade. This shouldn't be too much of a problem, it only means that after upgrade any users' browsers will have an invalid session and will need to re-login. That was already the case before this PR anyway since the Secret was never actually created. At least now the Secret is properly created and referenced in a better way.

elias-gb:
> @elias-gb could you test the latest commit I pushed and ensure it works on your setup?
> 
> ~The Secret you've defined should get properly created on installation still, but it will get recreated with a new secret value on upgrade. This shouldn't be too much of a problem, it only means that after upgrade any users' browsers will have an invalid session and will need to re-login. That was already the case before this PR anyway since the Secret was never actually created. At least now the Secret is properly created and referenced in a better way.~

Tested it now. Looks fine :)

charts/cryostat/templates/cookie_secret.yaml
```diff
@@ -0,0 +1,13 @@
...
+metadata:
+  name: {{ .Release.Name }}-cookie-secret
+type: Opaque
+{{ if $secret -}}
+data:
+  COOKIE_SECRET: {{ $secret.data.COOKIE_SECRET }}
+{{ else -}}
+stringData:
+  COOKIE_SECRET: {{ randAlphaNum 32 }}
+{{ end }}
```
tthvo, line 13: How about keeping the "cryostat.cookieSecret" and reference it here in the `data` field as in other secrets that helm creates?

```go
{{/*
Generate or retrieve a default value for cookieSecret.
*/}}
{{- define "cryostat.cookieSecret" -}}
{{- $secret := (lookup "v1" "Secret" .Release.Namespace (printf "%s-cookie-secret" .Release.Name)) -}}
{{- if $secret -}}
{{/*
   Use the current secret. Do not regenerate.
*/}}
{{- $secret.data.COOKIE_SECRET -}}
{{- else -}}
{{/*
    Generate a new secret.
*/}}
{{- randAlphaNum 32 | b64enc | quote -}}
{{- end }}
{{- end }}
```

tthvo (commented):
Tested both on k8s (minikube) and OpenShift Local. Looks good^^

ebaron (approved):
Looks good to me, thank you for your contribution @elias-gb!

Issue #187, opened 2024-09-06
[Bug] ArgoCD compatibility


grzesuav:
### Current Behavior

As for now, when deploying with ArgoCD make the app desyncronized -  see picture.
<img width="1112" alt="image" src="https://github.com/user-attachments/assets/91550789-ead5-4dad-ba1c-18b123cf14b5">

### Expected Behavior

No sync differences 

### Steps To Reproduce

Just deploy cryostat via argocd

### Environment

```Markdown
- OS: 
- Environment: 
- Version:
```

### Anything else?

I suspect this is because of some helm random function used to generate one time password

grzesuav:
Main problem is `lookup` helm function is not supported in argocd - https://github.com/cryostatio/cryostat-helm/blob/main/charts/cryostat/templates/_helpers.tpl#L104
https://github.com/argoproj/argo-cd/issues/5202

would be nice to set this values manually (i.e.have possibility to define secret name in values, similar like https://github.com/cryostatio/cryostat-helm/blob/main/charts/cryostat/templates/db_secret.yaml#L1 is handled)

grzesuav:
the same I guess with https://github.com/cryostatio/cryostat-helm/blob/main/charts/cryostat/templates/cookie_secret.yaml#L1

andrewazores:
I filed #188 above as a separate issue to track the feature of supplying all the auto-generated secret values manually - it's a good idea in general. If that ends up being everything required to fix ArgoCD compatibility then great, otherwise I leave this as a separate issue in case more is required after that work is done.

andrewazores:
Linking to a relevant comment from the discussion above: https://github.com/argoproj/argo-cd/issues/5202#issuecomment-834134792

This comment is describing exactly the same thing we do with the same motivation.

Other useful-looking links from that discussion:

- https://github.com/kuuji/helm-external-val
- https://github.com/jkroepke/helm-kubectl

----

On the other hand, from that same discussion, ArgoCD is trying to come up with a solution so that Helm charts like ours should be usable (ie let the `lookup` function work).

https://github.com/argoproj/argo-cd/issues/5202#issuecomment-1435383896

https://github.com/argoproj/argo-cd/issues/5202#issuecomment-1644432182

So there may be some solution on the Argo side as well, even if we don't get around to fully fixing this in our Helm chart first.

andrewazores:
[...]

PR #189, opened 2024-09-06
feat(secrets): enable user-supplied storage and authproxy secrets

### LINKED ISSUE #188: [Bug] All autogenerated secrets should be possible for user to manually supply
### Current Behavior

See https://github.com/cryostatio/cryostat-helm/issues/187

### Expected Behavior

_No response_

### Steps To Reproduce

_No response_

### Environment

_No response_

### Anything else?

_No response_

andrewazores:
Fixes #188


charts/cryostat/templates/_helpers.tpl
```diff
@@ -119,7 +119,7 @@ Get or generate a default secret key for object storage.
 Get or generate a default secret key for auth proxy cookies.
 */}}
 {{- define "cryostat.cookieSecret" -}}
-{{- $secret := (lookup "v1" "Secret" .Release.Namespace (printf "%s-cookie-secret" .Release.Name)) -}}
+{{- $secret := (lookup "v1" "Secret" .Release.Namespace ( default (printf "%s-cookie-secret" .Release.Name) .Values.authentication.cookieSecretName )) -}}
```
tthvo, line 122: I think these secret templates are only used with generated secrets (i.e.if provided secrets are not used) so we can leave them as they were previously?

That probably avoid misunderstanding that provided secrets are being looked up here.

charts/cryostat/templates/storage_secret.yaml
```diff
@@ -1,7 +1,9 @@
+{{- if empty .Values.storage.storageSecretName -}}
 apiVersion: v1
 kind: Secret
 metadata:
-  name: {{ .Release.Name }}-storage
+  name: {{ .Release.Name }}-storage-secret
 type: Opaque
 data:
-  SECRET_KEY: {{ include "cryostat.objectStorageSecretKey" . }}
+  STORAGE_ACCESS_KEY: {{ include "cryostat.objectStorageSecretKey" . }}
```
tthvo, line 8: Just a question tho: Changing the secret name or secret key name here means when upgrading to new chart version, this would break the existing connection to S3 storage instance?

tthvo: I guess if its not used for encrypting data in S3, then it should be fine?

andrewazores: Yes, it would. This next release is billed as a new major version (4.0) with other breaking changes anyway. Hopefully after this one we will have only minor semver release versions and nice upgrade paths for quite a long time.

tthvo: Oh right sounds good! Thanks!!

andrewazores: Right, this secret isn't used for encrypting the data at rest in S3, it's basically the service account password for authenticating to S3.

tthvo: ohh then it should not break the connection the. I haven't checked tho but will keep an eye for that :D

tthvo:
I guess we are waiting on response in #187? Maybe we can move forward and when the all changes are merged on `main`, anyone can pull the latest OCI image?

andrewazores:
Yes, I say we go ahead. It would be easier for Grzegorz to test anyway if it was merged, I think, and if there are any other fixes required we can do them as separate PRs.

tthvo:
Hmm some issues during rebasing? Some conflicts seem to be reported and the change diffs are a bit odd :D

andrewazores:
Ah, I think I just rebased it onto main instead of the separate-db-storage branch.

charts/cryostat/templates/storage_secret.yaml
```diff
@@ -1,7 +1,9 @@
+{{- if empty .Values.storage.storageSecretName -}}
 apiVersion: v1
 kind: Secret
 metadata:
-  name: {{ .Release.Name }}-storage
+  name: {{ .Release.Name }}-storage-secret
```
tthvo, line 5: I think Secrets should have the common labels, similar to other resource. This applies to all generated secrets.

```
  labels:
    {{- include "cryostat.labels" $ | nindent 4 }}
```

charts/cryostat/templates/cryostat_deployment.yaml
```diff
@@ -84,8 +84,8 @@ spec:
           - name: QUARKUS_S3_AWS_CREDENTIALS_STATIC_PROVIDER_SECRET_ACCESS_KEY
```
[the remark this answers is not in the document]
andrewazores: Good catch. I added this to the base `separate-db-storage` branch.

tthvo (approved):
Looks good to me :D

PR #196, opened 2024-09-06
test(scripts): add scripts to ease local chart testing

andrewazores:
- **test(ct): add utility script for running ct locally**
- **add script for running connection tests**

### DIFF
--- a/ct.bash
+++ b/ct.bash
@@ -2,8 +2,18 @@
 
 set -xe
 
+if ! command -v ct; then
+    echo "ct command not found in \$PATH"
+    exit 1
+fi
+
 TEST_NAMESPACE="${TEST_NAMESPACE:-helm-test}"
 CLUSTER_CLIENT="${CLUSTER_CLIENT:-$(which kubectl)}"
+
+if ! command -v "${CLUSTER_CLIENT}"; then
+    echo "${CLUSTER_CLIENT} command not found in \$PATH"
+    exit 1
+fi
 
 function cleanup() {
     "${CLUSTER_CLIENT}" delete ns "${TEST_NAMESPACE}"
--- a/ct.bash
+++ b/ct.bash
@@ -1,0 +1,13 @@
+#!/usr/bin/env bash
+
+set -xe
+
+TEST_NAMESPACE="${TEST_NAMESPACE:-helm-test}"
+
+function cleanup() {
+    kubectl delete ns "${TEST_NAMESPACE}"
+}
+trap cleanup EXIT
+
+kubectl create ns "${TEST_NAMESPACE}"
+ct install --remote "${REMOTE:-origin}" --target-branch "${TARGET_BRANCH:-main}" --namespace="${TEST_NAMESPACE}" --config ct.yaml --debug "$@"
--- a/ct.bash
+++ b/ct.bash
@@ -3,11 +3,12 @@
 set -xe
 
 TEST_NAMESPACE="${TEST_NAMESPACE:-helm-test}"
+CLUSTER_CLIENT="${CLUSTER_CLIENT:-$(which kubectl)}"
 
 function cleanup() {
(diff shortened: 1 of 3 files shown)


tthvo (commented):
Make sense to convert these into a `Makefile`?

andrewazores:
I could, but I think it would mostly end up being just inline shell commands anyway, so I don't know how much that would really gain.

tthvo:
> I could, but I think it would mostly end up being just inline shell commands anyway, so I don't know how much that would really gain.

Ah okayy! I was just thinking Makefile could define some targets that install the chart for some scenarios (i.e. port-forward, with route, with ingress, etc) to help with testing.

andrewazores:
It could, but unless there is a particular reason to use a Makefile, I'd rather not :-) if it adds something useful on top of bash scripts then yes, or else if it's also going to provide conventional `make clean install` kind of commands for users/developers. But if it's only for running different test scenarios then that sounds no better (and IMO, worse) than just having some bash scripts.

tthvo:
> It could, but unless there is a particular reason to use a Makefile, I'd rather not :-) if it adds something useful on top of bash scripts then yes, or else if it's also going to provide conventional `make clean install` kind of commands for users/developers. But if it's only for running different test scenarios then that sounds no better (and IMO, worse) than just having some bash scripts.

Oh alright! I got no specific reasons to use Makefile (just used to have them around). Otherwise, bash scripts sound good to me :D

ct.bash
```diff
@@ -0,0 +1,13 @@
+#!/usr/bin/env bash
+
+set -xe
+
+TEST_NAMESPACE="${TEST_NAMESPACE:-helm-test}"
+
+function cleanup() {
+    kubectl delete ns "${TEST_NAMESPACE}"
```
tthvo, line 8: Also take into account `oc` maybe?

ct.bash
```diff
@@ -0,0 +1,13 @@
...
+}
+trap cleanup EXIT
+
+kubectl create ns "${TEST_NAMESPACE}"
+ct install --remote "${REMOTE:-origin}" --target-branch "${TARGET_BRANCH:-main}" --namespace="${TEST_NAMESPACE}" --config ct.yaml --debug "$@"
```
tthvo, line 13: Maybe a safe check if `ct` is available?

k8s-test.bash
```diff
@@ -0,0 +1,23 @@
...
+fi
+
+kubectl create ns "${TEST_NAMESPACE}"
+function cleanup() {
+    kubectl delete ns "${TEST_NAMESPACE}"
+    if [ "${CREATE_CLUSTER:-true}" = "true" ]; then
+        kind delete cluster
+    fi
+}
+trap cleanup EXIT
```
tthvo, line 19: Maybe this part can come before setting up the cluster? Just in case some error in the setup that won't be cleaned up.

k8s-test.bash
```diff
@@ -0,0 +1,44 @@
...
+RELEASE_NAME="${RELEASE_NAME:-cryostat-k8s-test}"
+
+function cleanup() {
+    "${CLUSTER_CLIENT}" delete ns "${TEST_NAMESPACE}"
+    if [ "${CREATE_CLUSTER:-true}" = "true" ]; then
+        kind delete cluster
+    fi
+}
+trap cleanup EXIT
+cleanup
```
tthvo, line 27: Maybe not `cleanup` here? When I don't have a cluster running , the command `kubectl delete` will hang for quite sometimes (i.e. trying to contact an old cluster I have elsewhere that is still in kubeconfig).

andrewazores: Ah, fair enough. It doesn't hang for very long for me, but I guess it depends on the existing configurations.

tthvo (approved):
Looks good^^

PR #192, opened 2024-09-06
feat(deployments): separate db, storage, and reports deployments

andrewazores:
Fixes #114 #185 #188

Includes #184 #189 #191

Issue #200, opened 2024-09-23
[Request] Include semver major number in resource names


andrewazores:
This is just another temporary/one-time workaround the same as `--force` IMO, because we would have to do another rename again in the future if we change the labels again or do anything else similar.

Maybe a naming scheme like `{{ include "cryostat.fullname" . }}-v{{ cryostat.semver.major }}`, which would expand out to ex. `cryostat-v3`, could be usable across the various resources. This way their names stay stable within a major release version, but on major version upgrade when things are expected to change in a breaking way anyway, then the resources are renamed and upgrades can be done without force.

_Originally posted by @andrewazores in https://github.com/cryostatio/cryostat-helm/pull/197#discussion_r1772039428_

tthvo:
https://helm.sh/docs/chart_template_guide/function_list/#semver

andrewazores:
Done in #202

PR #201, opened 2024-09-23
ci(test): add upgrade strategy matrix fort helm tests

tthvo:
Related to #192 

## Description of changes

Added a matrix to specify upgrade strategy and whether to force upgrades. I think this would help us test upgrade path with any supported strategies.


charts/cryostat/templates/cryostat_deployment.yaml
```diff
@@ -3,7 +3,7 @@
 apiVersion: apps/v1
 kind: Deployment
 metadata:
-  name: {{ include "cryostat.fullname" . }}
+  name: {{ include "cryostat.fullname" . }}-main
```
tthvo, line 6: I added this change so that `ct` can detect changes and run tests for this PR :)) 

If possible, I think it could be a temporary fix until #200.  Preferably, we might want to avoid using '--force' in upgrades.

tthvo:
Sorry for the noises from CI fails :D I have been validating the behaviour of `helm upgrade --force`, but I misunderstood it for the helm v2 behaviour.

In helm v3, `helm upgrade --force` will work similarly to `kubectl replace`, which will fail on immutable field changes. I think we do need #200 for the upgrade to work at all.

tthvo:
> In helm v3, helm upgrade --force will work similarly to kubectl replace, which will still fail on immutable field changes. I think we do need https://github.com/cryostatio/cryostat-helm/issues/200 for the upgrade to work at all.

I guess #200 would help allow a nice and seamless upgrade for users. Though, there are other ways to upgrade that requires a bit of work. Here are some ways I tested that worked.

**1. Delete the release but keep the release history & install with replace**

```bash
$ helm uninstall --keep-history <release-name>
$ helm install --replace <release-name> <chart-source>
```

All resources of previous release is removed but the release history is kept intact. However, rollback would fail (i.e. immutable field).

**2. Delete the main deployment & upgrade**

```bash
kubectl -n <release-namespace> delete deployment/<main-deployment>
helm upgrade <release-name> <chart-source>
```

Since the deployment is deleted, helm will recreate it when upgrading. Again, rollback would fail unless the current deployment is deleted. 

I came across this in the Prometheus chart: https://github.com/prometheus-community/helm-charts/blob/main/charts/prometheus/README.md#to-220

**3. Rename the deployment**

This is the idea of #200 but I guess it is simpler since we can append `-main` to the main deployment. Rollback should work as the deployment is recreated.

All above ways are, like you said, one time temporary workaround. #200 is still nicest way around :D

andrewazores (approved):

Issue #168, opened 2024-07-09
[Epic] TLS configuration enhancements


andrewazores:
### Describe the feature

End users should have the following new options for TLS:

1. If using OpenShift, enable the `serving-cert` feature. This is implemented now, but only in a way where it is tied to deployment of the `openshift-oauth-proxy`
2. Supply their own custom certs
3. Configure the auth proxy (OpenShift or OAuth2) to use custom certs
4. Auto-configure the auth proxy (OpenShift or OAuth2) to use OpenShift `serving-cert`, if `serving-cert` is enabled and no custom certs are supplied

andrewazores:
Some discussion here: https://github.com/cryostatio/cryostat-helm/pull/167#issuecomment-2218643815

tthvo:
I guess there is one small thing to note is that the oauth proxy seems not to set some `X-Forwarded-*` header so redirect will likely fail. When ingress or route is available, those headers are set and forwarded correctly.

https://github.com/mwangggg/cryostat3/blob/35f8a9eff8a3080d2c004ac65efab6c2749ac2f3/compose/auth_proxy_alpha_config_https.yaml#L29-L35

andrewazores:
The proxy seems not to set those headers on its own, but that configuration adds them in so that it does set them. So long as the relevant environment variables get set then it should work out, I think.

andrewazores:
https://github.com/cryostatio/cryostat-helm/issues/115#issuecomment-2445174467

PR #206, opened 2024-10-31
feat(reports): add auth proxy protecting report generators

### LINKED ISSUE #204: [Request] Auth for reports sidecars
### Describe the feature

References:
https://quarkus.io/guides/rest-client#default-header-factory
https://quarkus.io/guides/security-basic-authentication-howto

The reports generator deployment is currently unsecured, so anyone within the cluster can send requests to its Service. There should be authentication on this.

When Cryostat is deployed with the oauth2 proxy for Basic authentication then the same proxy could also be placed in front of the Reports containers. It could share the same credentials configuration so that the same users can have direct access, or else it could have separate generated credentials in a Secret so that only Cryostat can authenticate against the Reports proxy (similar to a serviceaccount).

It is also possible to simply configure the Reports containers to enable Basic authentication directly, rather than placing an oauth2 proxy in front. This could again either reuse the same users configuration, or else it could have unique credentials that are only known to Cryostat.

When deployed on OpenShift and using the OpenShift OAuth proxy then it also seems reasonable to reuse the OpenShift OAuth proxy here, allowing Cryostat's serviceaccount. This also has the automatic TLS configuration benefit, so it should be preferred when it is feasible.

_No response_

### LINKED ISSUE #207: [Bug] Remove documentation about oauth2-proxy workaround for default namespace deployment
I guess after this is resolved, we can also remove the note here about setting `runAsUser` in default namespace  (i.e. The `oauth2-proxy` container image also has `USER` set):

https://github.com/cryostatio/cryostat-helm/blob/5bda8443d8a41b8f259aae607e88054edce92992/charts/cryostat/values.yaml#L276

_Originally posted by @tthvo in https://github.com/cryostatio/cryostat-operator/issues/969#issuecomment-2458427670_

andrewazores:
Fixes #204
Fixes #207

1. Adds an auth proxy container to each report replica Pod. This is oauth2-proxy normally, or else openshift-oauth-proxy if OpenShift authentication is enabled
2. Regardless of which proxy container is deployed, it is configured to only use Basic auth
3. A Secret containing a randomly generated access key and a corresponding htpasswd file is created. The htpasswd file is loaded by the auth proxy to set up its Basic auth. The access key is referenced by Cryostat so that it can be used for Basic authentication on Cryostat's internal REST client which it uses to send requests to the report generator service.
4. If the proxy is an openshift-oauth-proxy, then the automatic OpenShift TLS configuration is picked up and the Service points at the proxy's HTTPS port. Cryostat is configured to trust all certificates and not verify hostnames in this case (see #168). If the proxy is oauth2-proxy then the communications are done over unencrypted HTTP.

The end result should be that, if a report generator deployment is created (`--set reports.replicas=n`, where `n>0`), then requests sent to these Pods or the Service in front of them will require the client to pass an HTTP Basic authentication check. This check uses a hardcoded username, but a generated secret for a password. This helps prevent abuse of the report generator by other cluster workloads asking the report generator to process JFR files when they are not part of the Cryostat deployment.

To test:
1. Install the Helm chart as usual, but with `--set reports.replicas=1` to configure a report generator deployment.
2. Visit the Cryostat UI, create a sample application or define the `localhost:0` custom target, and request an automated analysis report
3. Verify that the report is generated successfully
4. Inspect the Pod logs and verify that the report generator Pod serviced the request
5. Launch a shell within the cluster (via OpenShift UI, or `kubectl exec` on ex. the Cryostat container) and do `curl -vk http://cryostat-reports.mynamespace.svc:10001/report`. Change the scheme to `https` if deployed with OpenShift authentication enabled. This should result in an HTTP 403.


charts/cryostat/templates/_reports_authproxy.tpl
```diff
@@ -0,0 +1,92 @@
...
+    - --cookie-secret=$(COOKIE_SECRET)
+    - --request-logging=true
+    - --openshift-service-account={{ include "cryostat.serviceAccountName" . }}
+    - --proxy-websockets=true
+    - --http-address=0.0.0.0:4180
+    - --https-address=:8443
+    - --tls-cert=/etc/tls/private/tls.crt
+    - --tls-key=/etc/tls/private/tls.key
+    - --proxy-prefix=/oauth2
+    - --bypass-auth-for=^/health?$
```
ebaron, line 29: Is the `?` supposed to be there?

ebaron (approved):

PR #210, opened 2024-11-14
feat(oauth2proxy): enable self-signed TLS cert

andrewazores:
Related to #168

Uses a Helm function to generate a self-signed cert, which is used for TLS on the oauth2-proxy. In the case that `authentication.openshift.enabled=true` is set, then the openshift-oauth-proxy is used instead and is presumably deployed on OpenShift, which has the *serving-cert* feature instead. If that value is not set, then the feature in this PR enables similar behaviour for the oauth2-proxy (whether deployed on OpenShift or not) so that the proxy only exposes HTTPS.


charts/cryostat/templates/NOTES.txt
```diff
@@ -59,6 +63,6 @@
 {{- else if contains "LoadBalancer" .Values.core.service.type }}
   echo http://$SERVICE_IP:{{ .Values.core.service.httpPort }}
 {{- else if contains "ClusterIP" .Values.core.service.type }}
-  http://localhost:8080
+  {{ ternary "https" "http" (or .Values.authentication.openshift.enabled .Values.oauth2Proxy.service.tls.selfSigned.enabled) }}://localhost:{{ ternary "8443" "8080" (or .Values.authentication.openshift.enabled .Values.oauth2Proxy.service.tls.selfSigned.enabled) }}
```
tthvo, line 66: Since `(or .Values.authentication.openshift.enabled .Values.oauth2Proxy.service.tls.selfSigned.enabled)` is being in many places, maybe we could refactor it into a helper named template?

charts/cryostat/templates/alpha_config.yaml
```diff
@@ -12,6 +12,14 @@ data:
   alpha_config.yaml: |-
     server:
       BindAddress: http://0.0.0.0:4180
+      {{- if .Values.oauth2Proxy.service.tls.selfSigned.enabled }}
+      SecureBindAddress: https://0.0.0.0:8443
+      {{- end}}
+      TLS:
```
tthvo, line 18: Should this be wrapped in the if block above?

charts/cryostat/values.yaml
```diff
@@ -267,6 +267,11 @@ oauth2Proxy:
     pullPolicy: Always
     ## @param oauth2Proxy.image.tag Tag for the OAuth2 Proxy container image
     tag: "latest"
+  service:
+    tls:
+      selfSigned:
+        ## @param oauth2Proxy.service.tls.selfSigned.enabled Whether a self-signed TLS certificate for oauth2-proxy HTTPS is generated and used.
+        enabled: false
```
tthvo, line 274: Could it be like below (i.e. removing `service:`)?

```yaml
    tls:
      selfSigned:
        ## @param oauth2Proxy.service.tls.selfSigned.enabled Whether a self-signed TLS certificate for oauth2-proxy HTTPS is generated and used.
        enabled: false
```

charts/cryostat/templates/tests/test-grafana-connection.yaml
```diff
@@ -5,6 +5,7 @@ metadata:
   labels:
     {{- include "cryostat.labels" . | nindent 4 }}
     app.kubernetes.io/component: test-grafana-connection
+    helm-test: cryostat
```
tthvo, line 8: Just curious, are there some reasons to add this label?

charts/cryostat/templates/cryostat_tls_secret.yaml
```diff
@@ -0,0 +1,15 @@
+{{- if (and (not (.Values.authentication.openshift).enabled) (.Values.oauth2Proxy.service.tls.selfSigned.enabled)) }}
+{{- $fullName := include "cryostat.fullname" . }}
+{{- $cert := genSelfSignedCert $fullName nil nil 365 }}
```
tthvo, line 3: Just a question: This means the certificate expires after a year right? This means the users have to rotate the certificate themselves?

Any thoughts about depending on `cert-manager` to manage certificates  like on the operator side?

charts/cryostat/templates/NOTES.txt
andrewazores: I've tried this, but I just get complaints from Helm that the `include`d template has type `string` instead of `bool` :/

charts/cryostat/templates/tests/test-grafana-connection.yaml
andrewazores: Just to make cleanup of the test Pods easier.

charts/cryostat/templates/cryostat_tls_secret.yaml
andrewazores: Yea, it'll expire after a year. Users have to rotate it themselves or figure out something else automated. I don't really want to get into adding an external dependency to the Helm chart, since part of the reason for the chart to exist is for users who can't use the Operator, ex. because they don't have full admin control of the cluster or whatever else. If the user needs TLS but can't install the Operator, then this gives them at least something to work with as a starting point, and then they can build whatever other automation they need on top to suit their particular deployment environment.

andrewazores:
Working to rebase atop #206 now.

charts/cryostat/templates/NOTES.txt
tthvo: > I've tried this, but I just get complaints from Helm that the included template has type string instead of bool :/

I guess we can return string `"true"`/`"false"` and do string compare? Or perhaps just return the protocol `https`/`http`? What do you think?

charts/cryostat/templates/tests/test-grafana-connection.yaml
tthvo: Sure, how about using having standard format for cryostat-specific label? Like below or similar options:

```
charts.cryostat.io/role: helm-test
```

charts/cryostat/templates/cryostat_tls_secret.yaml
tthvo: Sounds good! Thanks for explaining! I think a quick note in README about this would be helpful :D

ebaron (approved):

PR #211, opened 2024-11-15
feat(labels): add part-of selector label

andrewazores:
- **feat(labels): add part-of selector label**
- **chore(service): rename ports to remove 'cryostat-' prefix**

Related to https://github.com/cryostatio/cryostat-openshift-console-plugin/issues/2

### DIFF
--- a/charts/cryostat/templates/_helpers.tpl
+++ b/charts/cryostat/templates/_helpers.tpl
@@ -48,6 +48,7 @@ Selector labels.
 {{- define "cryostat.selectorLabels" -}}
 app.kubernetes.io/name: {{ include "cryostat.name" . }}
 app.kubernetes.io/instance: {{ .Release.Name }}
+app.kubernetes.io/part-of: cryostat
 {{- end }}
 
 {{/*
--- a/charts/cryostat/templates/cryostat_service.yaml
+++ b/charts/cryostat/templates/cryostat_service.yaml
@@ -18,11 +18,13 @@ spec:
       targetPort: 4180
       protocol: TCP
       name: http
+      appProtocol: http
     {{- if (.Values.authentication.openshift).enabled }}
     - port: 443
       targetPort: 8443
       protocol: TCP
       name: https
+      appProtocol: https
     {{- end }}
   selector:
     {{- include "cryostat.selectorLabels" $ | nindent 4 }}
--- a/charts/cryostat/templates/cryostat_service.yaml
+++ b/charts/cryostat/templates/cryostat_service.yaml
@@ -17,12 +17,12 @@ spec:
     - port: {{ .Values.core.service.httpPort }}
       targetPort: 4180
       protocol: TCP
-      name: cryostat-http
+      name: http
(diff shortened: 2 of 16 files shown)


andrewazores:
The `ct.bash --upgrade` tests fail with:

```
>>> helm upgrade cryostat-8sjfi4ytca charts/cryostat --namespace helm-test --reuse-values --wait --timeout=600s
Error: UPGRADE FAILED: cannot patch "cryostat-8sjfi4ytca-v4" with kind Deployment: Deployment.apps "cryostat-8sjfi4ytca-v4" is invalid: spec.selector: Invalid value: v1.LabelSelector{MatchLabels:map[string]string{"app.kubernetes.io/component":"cryostat", "app.kubernetes.io/instance":"cryostat-8sjfi4ytca", "app.kubernetes.io/name":"cryostat", "app.kubernetes.io/part-of":"cryostat"}, MatchExpressions:[]v1.LabelSelectorRequirement(nil)}: field is immutable && cannot patch "cryostat-8sjfi4ytca-v4-db" with kind Deployment: Deployment.apps "cryostat-8sjfi4ytca-v4-db" is invalid: spec.selector: Invalid value: v1.LabelSelector{MatchLabels:map[string]string{"app.kubernetes.io/component":"db", "app.kubernetes.io/instance":"cryostat-8sjfi4ytca", "app.kubernetes.io/name":"cryostat", "app.kubernetes.io/part-of":"cryostat"}, MatchExpressions:[]v1.LabelSelectorRequirement(nil)}: field is immutable && cannot patch "cryostat-8sjfi4ytca-v4-storage" with kind Deployment: Deployment.apps "cryostat-8sjfi4ytca-v4-storage" is invalid: spec.selector: Invalid value: v1.LabelSelector{MatchLabels:map[string]string{"app.kubernetes.io/component":"storage", "app.kubernetes.io/instance":"cryostat-8sjfi4ytca", "app.kubernetes.io/name":"cryostat", "app.kubernetes.io/part-of":"cryostat"}, MatchExpressions:[]v1.LabelSelectorRequirement(nil)}: field is immutable
```

tl;dr the label selectors are immutable, so upgrading the Deployment in-place can't happen. I think this is kind of failure is OK so long as it only happens on semver major versions when breakages are expected (see https://github.com/cryostatio/cryostat-helm/pull/202).

```bash
$ bash ct.bash 2>&1 | tee ct.log
$ bash ct.bash --upgrade 2>&1 | tee ct-upgrade.log
$ diff ct.log ct-upgrade.log
```

ebaron (commented):
We already have some unreleased label selector changes right? So I think the upgrade should still work after we renamed the deployment here like we did in the operator?

ebaron:
> ^ #197 and #202
> 
> I think a 3.0 -> 4.0 upgrade will work because of this renaming, but if I understand how the upgrade test works correctly, it's trying to upgrade from main to this PR - so essentially 4.0 to 4.0, and there is no renaming happening between versions, hence the apparent immutability conflict.

Yup that sounds totally fine to me.

ebaron (approved):

PR #215, opened 2024-11-27
feat(eventtemplates): declarative configuration of Event Template ConfigMaps

### LINKED ISSUE #214: [Task] Declarative Event Templates
### Describe the feature

See https://github.com/cryostatio/cryostat/issues/718 . This should be implemented on the Helm Chart since it is already done in Cryostat and the Operator.

_No response_

andrewazores:
Fixes #214
Related to https://github.com/cryostatio/cryostat-helm/issues/225
Depends on https://github.com/cryostatio/cryostat/pull/726

To test:
1. Create a configmap containing a JFR event template: `oc create configmap mytemplate --from-file=mytemplate.jfc`
2. `helm install --set core.image.repository=quay.io/andrewazores/cryostat --set core.image.tag=event-templates-declarative-1 cryostat --set core.config.eventTemplates.configMapNames='{mytemplate}' ./charts/cryostat`
3. Follow the post-install port forwarding steps, then open the Cryostat Web UI. Define a `localhost:0` custom target. Go to Events, then Event Templates, and ensure that the `mytemplate` custom Event Template is listed.


charts/cryostat/templates/cryostat_deployment.yaml
```diff
@@ -216,3 +222,8 @@ spec:
         secret:
           secretName: {{ .Release.Name }}-proxy-tls
       {{- end }}
+      {{- range .Values.core.config.eventTemplates.configMapNames}}
+      - name: {{ . }}
+        configMap:
+          name: {{ . }}
+      {{- end }}
```
tthvo, line 229: Any thoughts on using a projected volume here to combine all config-map contents into a single volume? One benefit is that we can define a single volumeMount on `/opt/cryostat.d/templates.d`, thus reducing the manifest/object def size.

Reference: https://kubernetes.io/docs/concepts/storage/projected-volumes/

charts/cryostat/values.yaml
```diff
@@ -85,6 +85,10 @@ core:
       builtInPortNumbersDisabled: false
       ## @param core.discovery.kubernetes.portNumbers [array] List of port numbers that the Cryostat application should look for in order to consider a target as JMX connectable
       portNumbers: []
+  config:
+    eventTemplates:
+      ## @param core.config.eventTemplates.configMapNames [array] List of ConfigMap names. Each ConfigMap is expected to contain one or more files, which are .jfc (XML) JFR Event Templates, to be mounted to the Cryostat container.
+      configMapNames: []
```
tthvo, line 91: Wondering if we can allow specifying the key(s) within the configmaps? Something similar to [Cryostat CR](https://github.com/cryostatio/cryostat-operator/blob/0ec9a84ffded9865671402790379c2884097f456/api/v1beta2/cryostat_types.go#L493). 

I guess one case could be that a user uses a Configmap to put all Cryostat configurations (i.e. Event template and Automated Rules). I might be overthinking :D

charts/cryostat/templates/cryostat_deployment.yaml
```diff
@@ -216,3 +222,8 @@ spec:
         secret:
           secretName: {{ .Release.Name }}-proxy-tls
       {{- end }}
+      {{- range .Values.core.config.eventTemplates.configMapNames}}
+      - name: {{ . }}
+        configMap:
```
tthvo, line 227: Also, should we enforce that these configmaps are available at the time of installation (i.e. `optional: false`)?

charts/cryostat/values.yaml
andrewazores: I thought about that, but it seems like it'd be pretty messy trying to put all of the configurations of different types into one ConfigMap and use filenames to differentiate them. It would also much more quickly run into the ConfigMap size limit that way. I think using one or more ConfigMaps for each type of configuration file is a good medium option, where if there are many custom Event Templates that add up to too large a filesize then they can be split across multiple ConfigMaps, for example, but otherwise they can just be combined together as separate files and mounted that way.

charts/cryostat/templates/cryostat_deployment.yaml
andrewazores: That's a fair point.

andrewazores: From what I can tell, I think `optional: false` is the default (which makes sense), but I can mark them as explicitly non-optional.

charts/cryostat/templates/cryostat_deployment.yaml
tthvo: hm, I suppose reducing the number of volume mount definitions can help reduce the manifest size (thus, object size in etcd). And grouping them under a single volume seems to make it easier to recognize their purpose when reading the manifest (my opinion).

charts/cryostat/values.yaml
tthvo: Sure, that makes sense! I think the current way is good then^^

charts/cryostat/templates/cryostat_deployment.yaml
tthvo: I guess we can do the same for https://github.com/cryostatio/cryostat-helm/pull/219.

charts/cryostat/templates/cryostat_deployment.yaml
```diff
@@ -216,3 +222,11 @@ spec:
         secret:
           secretName: {{ .Release.Name }}-proxy-tls
       {{- end }}
+      - name: declarative-event-templates
```
tthvo, line 225: Should we set the `defaultMode` to read-only (i.e. `0440`)?

charts/cryostat/templates/cryostat_deployment.yaml
tthvo: Oh right, sounds good!

Josh-Matsuoka (approved):

ebaron (dismissed):

Issue #220, opened 2024-11-29
[Bug] Split Deployments cleanup


andrewazores:
### Current Behavior

Since https://github.com/cryostatio/cryostat-helm/pull/192 there are some peripheral configuration pieces that are still shared across the Deployments, but should not be, particularly in the chart's "other parameters":

https://github.com/cryostatio/cryostat-helm/tree/b91e9bd3c3e0958725290eaf06c5271bb3096ef1?tab=readme-ov-file#other-parameters

- at the least, some of these descriptions should be updated to reflect that they affect the whole installation, not only the "Cryostat deployment" (which sounds like the one Deployment for Cryostat itself, and not the database, storage, etc.)
- the `serviceaccount` that gets created is currently shared across all Deployments. Only Cryostat itself really needs any of the additional `role`s, so only it should use the `serviceaccount`. The other Deployments should be OK to use their default service accounts.
- `podAnnotations` should be distributed across each of the deployments, so that custom annotations can be specified for any of them. It may make sense to keep a global `podAnnotations` parameter which applies to all of them, so that common annotations can be placed here rather than repeated everywhere.
- `podSecurityContext` should probably have individual settings as well as a global shared setting to default to.
- `nodeSelector` should also have individual settings and a global fallback/default.
- `tolerations` same as above
- `affinity` same as above
- `pvc.*` should be moved to the Deployments which are expected to actually need persistent claims, ie. storage and database, and should be separated so that they can have different size requests or storage classes etc.

PR #221, opened 2024-11-29
fix(serviceaccount): non-Cryostat Pods run as default serviceaccount

andrewazores:
Related to https://github.com/cryostatio/cryostat-helm/issues/220

To test:
1. `helm install cryostat ./charts/cryostat`
2. Follow post-install port-forwarding steps
3. Deploy another sample application.
4. Open Cryostat Web UI and click around. Everything should work as normal. Sample application should be discovered.


tthvo (approved):
Looks good to me :D

ebaron (dismissed):

PR #224, opened 2024-11-29
fix(pvc): split PVC configuration between Deployments

andrewazores:
Related to #220

Splits PVC definition in two. Originally, storage only needed to be provided to the Cryostat container itself, for the archives. Later we added the h2:file database in 2.x, which was still part of the Cryostat container. Since Cryostat 3.0 we have a separate database and a separate S3 storage running as separate containers in the same Pod. In 4.0 we have separate Deployments for each of Cryostat, database, and storage. Therefore, it only makes sense that the database and storage have storage configuration parameters (and Cryostat does not), and both of these should also be configured separately since the size and speed requirements for each are likely to differ significantly.


charts/cryostat/templates/db_deployment.yaml
```diff
@@ -85,12 +85,12 @@ spec:
         {{- toYaml . | nindent 8 }}
       {{- end }}
       volumes:
-      {{- if ((.Values.pvc).enabled) }}
+      {{- if ((.Values.db.resources.pvc).enabled) }}
```
tthvo, line 88:
```suggestion
      {{- if (.Values.db.pvc).enabled }}
```

I think `.db.resources` refers to container resource requirements :D

charts/cryostat/templates/db_deployment.yaml
```diff
@@ -85,12 +85,12 @@ spec:
...
       - name: {{ .Chart.Name }}-db
         persistentVolumeClaim:
           claimName: {{ .Release.Name }}-db
       {{- end }}
-      {{- if not ((.Values.pvc).enabled) }}
+      {{- if not ((.Values.db.resources.pvc).enabled) }}
```
tthvo, line 93:
```suggestion
      {{- else }}
```

I guess we can use `else` here?

charts/cryostat/templates/storage_deployment.yaml
```diff
@@ -99,12 +99,12 @@ spec:
         {{- toYaml . | nindent 8 }}
       {{- end }}
       volumes:
-      {{- if ((.Values.pvc).enabled) }}
+      {{- if ((.Values.storage.resources.pvc).enabled) }}
```
tthvo, line 102:
```suggestion
      {{- if (.Values.storage.pvc).enabled }}
```

charts/cryostat/templates/storage_deployment.yaml
```diff
@@ -99,12 +99,12 @@ spec:
...
       - name: {{ .Chart.Name }}-storage
         persistentVolumeClaim:
           claimName: {{ .Release.Name }}-storage
       {{- end }}
-      {{- if not ((.Values.pvc).enabled) }}
+      {{- if not ((.Values.storage.resources.pvc).enabled) }}
```
tthvo, line 107: We can also use `else` here :D

charts/cryostat/templates/db_deployment.yaml
andrewazores: Ah right, thanks. I initially nested it under `resources`, but then it would be annoying to set PVC settings and CPU/memory separately so I moved it back up a level. I'll change this when I'm back at my computer (probably Monday).

charts/cryostat/values.yaml
```diff
@@ -180,6 +194,20 @@ storage:
...
+    annotations: {}
+    ## @param storage.pvc.storage Storage size to request for the persistentVolumeClaim
+    storage: 500Mi
+    ## @param storage.pvc.accessModes Access mode for the persistentVolumeClaim. See: [Access Modes](https://kubernetes.io/docs/concepts/storage/persistent-volumes/#persistentvolumeclaims)
+    accessModes:
+      - ReadWriteOnce
+    ## @param storage.pvc.selector [object] Selector for the persistentVolumeClaim. See: [Selector](https://kubernetes.io/docs/concepts/storage/persistent-volumes/#persistentvolumeclaims)
+    selector: {}
+    ## @param storage.pvc.storageClassName [string, nullable] The name of the StorageClass for the persistentVolumeClaim. See: [Class](https://kubernetes.io/docs/concepts/storage/persistent-volumes/#persistentvolumeclaims)
+    storageClassName: ""
```
tthvo, line 210: I think these `storageClassName` are meant default to `null` or unset (i.e. to be commented ou). This allows the default storageClass to to be set on the PVC.

https://kubernetes.io/docs/concepts/storage/persistent-volumes/#class-1

charts/cryostat/templates/db_deployment.yaml
tthvo: > I'll change this when I'm back at my computer (probably Monday).

Sounds good! Happy Black Friday weekends^^

ebaron: I think these still need `resources` removed from the path.

ebaron (approved):

PR #226, opened 2024-12-09
WIP EndpointSlices discovery

andrewazores:
Temporary PR to go along with https://github.com/cryostatio/cryostat/pull/740.

**THIS PR IS NOT MEANT TO BE MERGED**. This is only here for ease of deploying and testing the `EndpointSlices`-based discovery implementation. When that PR is ready on Cryostat itself, the old `Endpoints`-based discovery should be completely replaced as internal implementation details, reusing all the same config properties etc. The Helm chart should not actually need any updates to work with those internal changes.

### DIFF
--- a/charts/cryostat/templates/cryostat_deployment.yaml
+++ b/charts/cryostat/templates/cryostat_deployment.yaml
@@ -127,6 +127,18 @@ spec:
             value: {{ include "cryostat.commaSepList" (list .portNumbers 9091 .builtInPortNumbersDisabled) }}
           {{- end }}
           {{- end }}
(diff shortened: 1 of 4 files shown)


Josh-Matsuoka:
Seems like one of the workflows is still failing after re-running it, is that expected?

andrewazores:
I think it needs https://github.com/cryostatio/cryostat/pull/740 to go in first, since the test that's failing will actually deploy a Cryostat instance. This PR changes some permissions so without the corresponding Cryostat change I think it's going to fail at runtime due to the wrong permissions.

Josh-Matsuoka (approved):

Issue #203, opened 2024-10-23
[Request] Add ability to add additional env variable - 


grzesuav:
### Describe the feature

I would like to switch logging to JSON - https://quarkus.io/guides/logging#configuration so setting the env var - apparently is shoudl be done by `QUARKUS_LOG_CONSOLE_JSON`. It is not possible to add any right now

andrewazores:
You can install using the Helm chart and then manually edit the Deployment that it creates afterward. But I agree, it would be nice to be able to specify these things at installation time.

andrewazores:
I'm thinking of maybe adding something like this:

```yaml
core:
  config:
    extra:
      configMapName: ""
```

If the user creates a ConfigMap and installs with `--set core.config.extra.configMapName=my-cm`, then the `cryostat-deployment.yaml` template could check whether a ConfigMap name was provided and if it was, mount that to the Cryostat container environment variables. It'll be up to the user to make sure they aren't causing conflicts with other environment variables set by other top-level configuration options.

andrewazores:
^ we could do the same for the other deployments too quite easily, ex.

```yaml
reports:
  config:
    extra:
      configMapName: ""
```

Maybe we also need a way to specify per-container ConfigMaps, since at least the main Deployment actually contains Cryostat, jfr-datasource, Grafana dashboard, and an optional auth proxy. So far this plan would only allow for setting additional environment variables on the Cryostat container, or else setting environment variables that would be applied to every container in the Pod.

andrewazores:
@tthvo has the best proposal here, significantly expanding on what I had above:

https://github.com/cryostatio/cryostat-helm/pull/247#issuecomment-2885416216

PR #248, opened 2025-05-16
feat(envs): add configurations to include extra envs for storage container

tthvo:
Related to #203 

## Descriptions

Adding an field `storage.config.extra.envVars` and `storage.config.extra.envSources` to allow users to configure additional environment variables for the storage container.

This can be extended to other containers (i.e. core, database, etc).


andrewazores:
Looks great to me.

> This can be extended to other containers (i.e. core, database, etc).

What are your thoughts on how to handle multi-container Pods?

andrewazores:
Makes sense, but I do see one spot where it might get funny: the auth proxy container can be used in more than one Deployment. Already it can be present in both the main Cryostat Pod and the Reports Pod. I don't know what kind of additional configuration the user might want to set on this container, but with this structure it would only be possible to give both instances the same configurations, which might not be suitable.

(It would also be nice if there could be someway to reuse these snippets in different places and not copy-paste the same structures everywhere, but I think this is a Helm/yaml limitation)

tthvo:
> (It would also be nice if there could be someway to reuse these snippets in different places and not copy-paste the same structures everywhere, but I think this is a Helm/yaml limitation)

Unfortunately, we need to copy-paste :sweat_smile:  I guess we can _generate_ the `values.yaml` from a template file, let's say `values.yaml.tpl` with [go/template](https://pkg.go.dev/text/template). But that sounds a  bit over-engineering :D

> Makes sense, but I do see one spot where it might get funny: the auth proxy container can be used in more than one Deployment. Already it can be present in both the main Cryostat Pod and the Reports Pod. I don't know what kind of additional configuration the user might want to set on this container, but with this structure it would only be possible to give both instances the same configurations, which might not be suitable.

Right, totally unaware of that :sweat_smile: I guess we can do the following where we have a common settings and additional settings for specific pod use case. This is similar to how k8s defines [`SecurityContext`](https://kubernetes.io/docs/tasks/configure-pod-container/security-context/).

The settings with pod-specific context will take precedence. This works well because env var that are defined later takes precendence [[0]](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.27/#podspec-v1-core).

```yaml
oauth2Proxy:
  config:
    ## @extra oauth2Proxy.config.extra Extra configurations for the OAuth2 Proxy container image
    extra:
      ## @param oauth2Proxy.config.extra.envVars [array] Extra environment variables for the OAuth2 Proxy container image. See: [Define Environment Variables](https://kubernetes.io/docs/tasks/inject-data-application/define-environment-variable-container/)
      envVars: []
      ## @param oauth2Proxy.config.extra.envSources [array] Sources for extra variables for the OAuth2 Proxy container image. See: [Define Environment Variables From ConfigMap](https://kubernetes.io/docs/tasks/configure-pod-container/configure-pod-configmap/#configure-all-key-value-pairs-in-a-configmap-as-container-environment-variables)
      envSources: []
    context:
      core:
      ## @param oauth2Proxy.config.extra.core.envVars [array] Extra configurations for the OAuth2 Proxy container image in main pod
      extra:
        envVars: []
        envSources: []
      ## @param oauth2Proxy.config.extra.report.envVars [array] Extra configurations for the OAuth2 Proxy container image in report pod
      report:
        extra:
          envVars: []
          envSources: []
```

tthvo:
Or maybe this would be a better value settings

```yaml
oauth2Proxy:
  config:
    ## @extra oauth2Proxy.config.extra Extra configurations for the OAuth2 Proxy container
    extra:
      ## @param oauth2Proxy.config.extra.envVars [array] Extra environment variables for the OAuth2 Proxy container. See: [Define Environment Variables](https://kubernetes.io/docs/tasks/inject-data-application/define-environment-variable-container/)
      envVars: []
      ## @param oauth2Proxy.config.extra.envSources [array] Sources for extra variables for the OAuth2 Proxy container. See: [Define Environment Variables From ConfigMap](https://kubernetes.io/docs/tasks/configure-pod-container/configure-pod-configmap/#configure-all-key-value-pairs-in-a-configmap-as-container-environment-variables)
      envSources: []
      podContext:
        main:
            envVars: []
            envSources: []
        report:
          envVars: []
          envSources: []
```

andrewazores:
I like that second style better.

And I agree, generating the `values.yaml` from a template sounds like it's getting into over-engineering. It might be an idea worth keeping in mind in case the configuration keeps getting more complex, though, since we already do have a lot of duplicated template snippets across the various components.

tthvo:
yea haha I wish helm figures something out about this :D but sounds good! I'll circle back near end of week for follow-up PRs!

tthvo:
Hey @andrewazores, this should be ready for reviews :smile: 

Instead of PRs, I separate the changes in separate commits for easy reviews. All containers now support extra env vars and env sources. PTAL!

Issue #253, opened 2025-07-08
[Request] User configuration options to plug in their own psql database instance


andrewazores:
### Describe the feature

See https://github.com/cryostatio/cryostat-operator/issues/1134

PR #247, opened 2025-05-15
feat(storage): add configuration for external object storage provider

andrewazores:
See #246

**Rough draft/prototype at this stage.**

Allows the user to configure an alternate S3 object storage provider, rather than requiring the use of [cryostat-storage](https://github.com/cryostatio/cryostat-storage).

## Manual Testing

1. Set up an unmanaged S3-compatible object storage instance. In this case we will actually use `cryostat-storage` again, but deployed independently. This assumes an OpenShift cluster is available and `oc` is logged in.

<details>

<summary>
Deploy cryostat-storage
</summary>

```bash
$ oc new-project objectstorage
$ oc create secret generic s3cred --from-literal=STORAGE_ACCESS_KEY_ID=cryostat --from-literal=STORAGE_ACCESS_KEY=verySecretKey1 # set credentials to secure the S3 endpoint
$ oc create -f seaweed.yaml
```
`seaweed.yaml`
```yaml
---
apiVersion: apps/v1
kind: Deployment
metadata:
  generation: 1
  labels:
    app.kubernetes.io/component: seaweed
    app.kubernetes.io/instance: seaweed
    app.kubernetes.io/name: seaweed
    app.kubernetes.io/part-of: seaweed
  name: seaweed
spec:
  progressDeadlineSeconds: 600
  replicas: 1
  selector:
    matchLabels:
      app.kubernetes.io/component: seaweed
      app.kubernetes.io/instance: seaweed
      app.kubernetes.io/name: seaweed
      app.kubernetes.io/part-of: seaweed
  strategy:
    type: Recreate
  template:
    metadata:
      labels:
        app.kubernetes.io/component: seaweed
        app.kubernetes.io/instance: seaweed
        app.kubernetes.io/name: seaweed
        app.kubernetes.io/part-of: seaweed
    spec:
      containers:
      - env:
        - name: CRYOSTAT_BUCKETS
          value: archivedrecordings,archivedreports,eventtemplates,probes
        - name: CRYOSTAT_ACCESS_KEY
          valueFrom:
            secretKeyRef:
              key: STORAGE_ACCESS_KEY_ID
              name: s3cred
              optional: false
        - name: CRYOSTAT_SECRET_KEY
          valueFrom:
            secretKeyRef:
              key: STORAGE_ACCESS_KEY
              name: s3cred
              optional: false
        - name: DATA_DIR
          value: /data
        - name: IP_BIND
          value: 0.0.0.0
        - name: REST_ENCRYPTION_ENABLE
[...]

PR #254, opened 2025-07-08
feat(db): add configuration for external database provider

andrewazores:
See #247
Related to #253
Depends on cryostatio/cryostat#981

The external database still must have the `pgcrypto` extension enabled, even if no `encrypt.key` configuration parameter is explicitly set.

```sql
CREATE EXTENSION IF NOT EXISTS pgcrypto;
```

ex.

```
$ helm install \
  --set core.image.repository=quay.io/andrewazores/cryostat \
  --set core.image.tag=psql-no-encrypt-6 \
  --set core.discovery.kubernetes.enabled=true \
  --set core.databaseSecretName=neondb \
  --set db.provider.url='jdbc:postgresql://REDACTED.neon.tech/neondb?sslmode=require&channel_binding=require' cryostat ./charts/cryostat/
```

where the Secret named `neondb` (in the same namespace) contains:
```yaml
ENCRYPTION_KEY: ''
USERNAME: 'neondb_owner'
CONNECTION_KEY: REDACTED
```
The redacted connection key is the password for the `neondb_owner` account. This is specific to https://neon.tech , but configuration for any other database-as-a-service should be similar.

### DIFF
--- a/charts/cryostat/templates/_helpers.tpl
+++ b/charts/cryostat/templates/_helpers.tpl
@@ -87,8 +87,12 @@ Cryostat service port. 8443 if TLS is enabled, 8080 otherwise.
 Check if the database secret contains a username.
 */}}
 {{- define "cryostat.databaseSecretHasUsernameKey" -}}
-{{- $secret := (lookup "v1" "Secret" .Release.Namespace (printf "%s-db" .Release.Name)) -}}
-{{ not (empty ($secret.data).USERNAME) }}
+{{- if empty .Values.core.databaseSecretName -}}
+{{ "false" -}}
+{{- else -}}
+{{- $secret := (lookup "v1" "Secret" .Release.Namespace .Values.core.databaseSecretName ) -}}
+{{ not (empty (($secret).data).USERNAME) -}}
+{{- end -}}
 {{- end -}}
 
 {{/*
--- a/charts/cryostat/templates/cryostat_deployment.yaml
+++ b/charts/cryostat/templates/cryostat_deployment.yaml
@@ -81,7 +81,7 @@ spec:
             value: {{ printf "%s://cryostat:$(REPORTS_PASS_SECRET_KEY)@%s-reports.%s.svc:%d" (include "cryostat.core.service.scheme" .) $fullName $.Release.Namespace (int .Values.reports.service.httpPort) }}
           {{- end }}
           {{- $hasSecret := (include "cryostat.databaseSecretHasUsernameKey" . ) }}
-          {{- if $hasSecret | eq "true" }}
+          {{- if ($hasSecret | eq "true") }}
           - name: QUARKUS_DATASOURCE_USERNAME
             valueFrom:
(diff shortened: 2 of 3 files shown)


andrewazores:
This fails because the database secret is now expected to contain a `USERNAME` value. This will default to `cryostat` if the secret is generated by the Helm installation, which is actually the same as what the previous default was hardcoded to be. Upgrading an existing installation to this PR will fail however, since the `USERNAME` key won't be found in the old secret.

andrewazores:
Tested with a Supabase free tier account and it works just fine. Same kind of configuration as for Neon, just different parameters (username/password and connection URL).

ebaron (commented):
Is it possible to do something like default to "cryostat" for USERNAME if not present in the secret? Just wondering how feasible it would be to not make this break the upgrade path.

charts/cryostat/templates/db_secret.yaml
```diff
@@ -8,6 +8,7 @@ metadata:
 type: Opaque
 immutable: true
 data:
-  ENCRYPTION_KEY: {{ include "cryostat.databaseEncryptionKey" . }}
+  USERNAME: {{ include "cryostat.databaseUsernameKey" . }}
```
andrewazores: Still having the same upgrade problem with the Secret being modified - I guess the problem lies here. There needs to be some logic here to only insert the new `USERNAME` key if this is a new Secret being created, otherwise if it's an upgrade operation then leave it alone.

Or maybe:
- make the Secret(s) mutable
- change the Secret(s) name so that the 4.1 version doesn't update the 4.0 one and just replaces it with a different reference

We do already have the `_helpers.tpl` utility for `cryostat.deploymentName` which uses the chart's `AppVersion` and just extracts the semver major version, and applies that to the Deployment objects it creates. If this included the minor version as well, and applied to all resources instead of only the Deployments, then maybe it would also work as an upgrade path? It does mean there would be more churn on every minor version upgrade, and for Secrets in particular it would lead to regenerating new values if the user did not specify a reference to their own that they created.

ebaron: I meant don't touch the secret at all. Just check for the USERNAME key in the deployment template. If it doesn't exist, set the `QUARKUS_DATASOURCE_USERNAME`/`POSTGRESQL_USER` to "cryostat"

ebaron: Even if we remove `databaseUsernameKey` and do not add a `USERNAME` key to the secret?

ebaron: Like the secret would only ever have a `USERNAME` key if the user provides a custom database secret.

andrewazores: Ahh, I see what you mean. That sounds like it would work.

andrewazores:
The secret lookup function definitely isn't working as intended at the moment. Working on it.

ebaron (approved):
Seems to work well now

PR #302, opened 2026-05-06
chore(version): configure release branch for 4.2

jtolentino1:
Related to https://github.com/cryostatio/cryostat/issues/1493


andrewazores (dismissed):

ebaron (dismissed):
I don't think we want to do a version bump here. The Helm Chart 2.2.0 should correspond to Cryostat 4.2.0

andrewazores:
Ah yes, #301 has the correct version bump (prepping `main` to later become the 4.3.0 branch), but this PR should leave the version as it is (previously prepped to become 4.2.0)

andrewazores:
Actually, same comment as here too: https://github.com/cryostatio/cryostat-operator/pull/1300#discussion_r3196702878

andrewazores:
@jtolentino1 ping - the auth proxy tags should also be pinned here

andrewazores (approved):

